PR #78, opened 2020-09-22
Adding VM creation fail event

apluchik:
Adding first event logging; for VM creation failure.

### DIFF
--- a/controllers/azurestackhcivirtualmachine_controller.go
+++ b/controllers/azurestackhcivirtualmachine_controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/microsoft/cluster-api-provider-azurestackhci/cloud/scope"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


JocelynBerrendonner (approved):
I am not an expert on what the recorder does, so please check with Zach and Madhan, otherwise this looks good to me

JocelynBerrendonner:
Note: when adding folks to PR, don't forget to also add microsoft/moc-pr which should include everyone

pradipd:
What does this event look like and where does it go? E.g. is this event going to a log file? is it being uploaded to azure or some other cloud?

madhanrm:
it would be available as part of kubctl get events

madhanrm (approved):

pradipd:
got it. thanks!

zawachte-msft:
lgtm. @apluchik lets sync up and talk about status and phases.

madhanrm:
Are we adding more error events ?

apluchik:
@madhanrm Yep, I'll submit another PR to add more events now.

PR #91, opened 2020-10-22
Create VMs with shortened computer names for Windows OS compatibility

nwoodmsft:
Same PR that we had before (https://github.com/microsoft/cluster-api-provider-azurestackhci/pull/81) - bringing this back now that other components are ready for the change.


zawachte-msft:
lgtm

PR #101, opened 2020-12-16
bumping moc-sdk-for-go version to pick up fix for static ip

pradipd:


zawachte-msft:
don't merge this until @apluchik gives an okay. he warned me bumping up caph further would cause issues.

apluchik:
> don't merge this until @apluchik gives an okay. he warned me bumping up moc further would cause issues until he finished some cloudagent work.
This won't be affected by my change as this PR is bumping to v0.10.6-alpha.2.
The breaking change I was referring to was only in the latest moc-sdk (v0.10.6-alpha.4), where some different fields are used.

apluchik (approved):

pradipd:
Thanks.  I figured that out.  I only bumped the version to include the fix I need.

PR #106, opened 2021-02-25
Changes for bare-metal

cwize1:
Haven't been able to test these changes yet. Just opening a draft PR to check with people if these changes make sense.


cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +196,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+					AdminPassword:      virtualMachine.VirtualMachineProperties.OsProfile.AdminPassword,
+					CustomData:         virtualMachine.VirtualMachineProperties.OsProfile.CustomData,
+					LinuxConfiguration: virtualMachine.VirtualMachineProperties.OsProfile.LinuxConfiguration,
+				},
+				NetworkProfile: &compute.BareMetalMachineNetworkProfile{
+					NetworkInterfaces: vmNetworkInterfacesToBareMetal(virtualMachine.VirtualMachineProperties.NetworkProfile.NetworkInterfaces),
+				},
+				HardwareProfile: &compute.BareMetalMachineHardwareProfile{
+					MachineSize: &compute.BareMetalMachineSize{
+						CpuCount: virtualMachine.VirtualMachineProperties.HardwareProfile.CustomSize.CpuCount,
```
jiria, line 266: We probably should not override this, right? This should come from MOC.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +196,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+				SecurityProfile:   virtualMachine.VirtualMachineProperties.SecurityProfile,
+				Host:              virtualMachine.VirtualMachineProperties.Host,
+				ProvisioningState: virtualMachine.VirtualMachineProperties.ProvisioningState,
+				Statuses:          virtualMachine.VirtualMachineProperties.Statuses,
+			}
+
+			// Try to apply the update.
+			_, err := s.BareMetalClient.CreateOrUpdate(ctx, resourceGroup, *bareMetalMachine.Name, &bareMetalMachine)
+			if mocerrors.IsInvalidVersion(err) {
+				// Another entity claimed the machine. Keep searching.
```
jiria, line 280: Not necessarily, some other update could have happened. We might not want to give up, though your top level for loop will resolve that.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +196,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+				continue
+			} else if err != nil {
+				return nil, err
+			}
+
+			// Success!
+			return &bareMetalMachine, nil
+		}
```
jiria, line 288: Should we sleep for a bit here? Maybe not if an entity was found, but failed on version. But otherwise, probably busy looping might not be good idea.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +196,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+		bareMetalMachineCount = len(*bareMetalMachineList)
+
+		// Go through list and try to find an unused machine.
+		for _, bareMetalMachine := range *bareMetalMachineList {
+			if isBareMetalMachineInUse(&bareMetalMachine) {
+				// Machine is already in use. Keep searching.
+				continue
+			}
+
+			unusedFound = true
```
jiria, line 243: This does not seem to add any value at the moment.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +196,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+				},
+				OsProfile: &compute.BareMetalMachineOSProfile{
+					ComputerName:       virtualMachine.VirtualMachineProperties.OsProfile.ComputerName,
+					AdminUsername:      virtualMachine.VirtualMachineProperties.OsProfile.AdminUsername,
+					AdminPassword:      virtualMachine.VirtualMachineProperties.OsProfile.AdminPassword,
+					CustomData:         virtualMachine.VirtualMachineProperties.OsProfile.CustomData,
+					LinuxConfiguration: virtualMachine.VirtualMachineProperties.OsProfile.LinuxConfiguration,
+				},
+				NetworkProfile: &compute.BareMetalMachineNetworkProfile{
+					NetworkInterfaces: vmNetworkInterfacesToBareMetal(virtualMachine.VirtualMachineProperties.NetworkProfile.NetworkInterfaces),
```
jiria, line 262: To anybody who knows. Does CAPH+MOC have functionality to specify static IPs?

jiria (commented):
Looking good from my perspective (though I am also still learning here).

cloud/services/virtualmachines/virtualmachines.go
cwize1: Good call.

cloud/services/virtualmachines/virtualmachines.go
cwize1: Correct. Though I was aiming for a simplified comment. I will extend the comment to be more specific.

cloud/services/virtualmachines/virtualmachines.go
cwize1: Eh, if* this code ever has efficiency issues then we should probably change the cloud agent API to allow a bare-metal machine to be acquired instead of having the controller search for an unused machine. 

*when

cloud/services/virtualmachines/virtualmachines.go
cwize1: `unusedFound` is set to `false` when the `for` loop loops round.

cloud/services/virtualmachines/virtualmachines.go
cwize1: Also, from the comment below, you are misreading how `unusedFound` is used. `unusedFound` will prevent this from looping ad infinitum, except in the edge case where there is massive churn in the bare-metal machines.

cloud/services/virtualmachines/virtualmachines.go
jiria: I think I am still having problem understanding the logic. Would not the top level for loop only ever allow for a single iteration?

cwize1: Opps, my logic is broken. `unusedFound` should be set to `false` at the start of an iteration not at the end.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -315,7 +321,9 @@ func (s *Service) Delete(ctx context.Context, spec interface{}) error {
 	switch vmSpec.MachineType {
 	case infrav1.MachineTypeBareMetal:
 		klog.V(2).Infof("deleting bare-metal machine %s ", vmSpec.Name)
-		err = s.BareMetalClient.Delete(ctx, s.Scope.GetResourceGroup(), vmSpec.Name)
+		if vmSpec.ResourceName != "" {
+			err = s.BareMetalClient.Delete(ctx, s.Scope.GetResourceGroup(), vmSpec.ResourceName)
```
jiria, line 325: I wonder if this should be just update on BMM entity to clear it, as opposed to delete it altogether. But fine for the initial round.

cwize1: Yes, you are right.

api/v1alpha3/azurestackhcivirtualmachine_types.go
```diff
@@ -75,11 +75,11 @@ type AzureStackHCIVirtualMachineStatus struct {
 	// +optional
 	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
 
-	// ResourceName is the name resource used by the machine.
+	// BackingResourceName is the name resource used by the machine.
```
jiria, line 78: Should we call it MocRresourceName?

cwize1: I am fine with either.

hack/tools/go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	k8s.io/code-generator v0.18.0-alpha.2.0.20200130061103-7dfd5e9157ef
 	sigs.k8s.io/cluster-api/hack/tools v0.0.0-20200319204836-a97903fa1e7e
-	sigs.k8s.io/controller-tools v0.2.6
+	sigs.k8s.io/controller-tools v0.2.9
```
cwize1, line 24: FYI: I was getting weird build errors when trying to update the generated stuff. Bumping this to the latest patch version seemed to fix the problem.

config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhciclusters.yaml
```diff
@@ -249,41 +258,56 @@ spec:
...
               conditions:
-                description: Conditions defines current service state of the cluster.
+                description: Conditions defines current service state of the AzureStackHCICluster.
                 items:
-                  description: Condition defines an observation of a Cluster API resource operational state.
+                  description: Condition defines an observation of a Cluster API resource
+                    operational state.
                   properties:
                     lastTransitionTime:
-                      description: Last time the condition transitioned from one status to another. This should be when the underlying condition changed. If that is not known, then using the time when the API field changed is acceptable.
```
cwize1, line 257: FYI: These formatting changes are a result of updating the patch version of the `controller-tools` tool.

jiria (approved):

nwoodmsft:
Will be taking a look today, please do not merge until you have sign-off from either Zach or I.

api/v1alpha3/azurestackhcivirtualmachine_types.go
nwoodmsft: MOC isn't really mentioned in the CRDs, we usually expose up as AzureStackHci and keep MOC as more of an internal detail.

I'd suggest any of:
* ResourceName
* AzureStackHciResourceName
* CloudResourceName

api/v1alpha3/types.go
```diff
@@ -247,3 +247,15 @@ const (
 	// AzureStackHCIClusterPhaseUnknown is returned if the Cluster state cannot be determined.
 	AzureStackHCIClusterPhaseUnknown = AzureStackHCIClusterPhase("")
 )
+
+// HostType specifies what type of machine a node should be deployed on.
+type HostType string
+
+const (
+	// HostTypeVM specifies that the node should be deployed on a virtaul machine.
```
nwoodmsft, line 255: Nit: typo "virtaul"

hack/tools/go.mod
zawachte-msft: thanks. 0.2.9 is good.

zawachte-msft: imo controller-tools shouldn't even be in the go.mod. while controller-gen is needed to build the manifest, its not needed for the manager code.

cloud/services/virtualmachines/virtualmachines.go
nwoodmsft: Yes it does. We support static ip deployments in AKS HCI today (https://docs.microsoft.com/en-us/azure-stack/aks-hci/concepts-networking#virtual-networks). There is a property on network interface to specify a static ip and there is also a way on the vnet to mark the subnet as static so that all nics connected to that vnet receive static ip allocations from it.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -180,27 +203,138 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+			} else if err != nil {
+				return nil, err
+			}
+
+			// Success!
+			return &bareMetalMachine, nil
+		}
+	}
+
+	return nil, mocerrors.Wrapf(mocerrors.OutOfCapacity, "No free bare-metal nodes. Total nodes [%d]", bareMetalMachineCount)
```
nwoodmsft, line 295: Nit: I don't see value add from using mocerrors here as it just calls errors.Wrapf underneath. We directly use errors.Wrapf elsewhere in the file and can do that here too.

hack/tools/go.mod
cwize1: Note that this is the `/hack/tools/go.mod` file, not the `/go.mod` file. Also, the `Makefile` uses this to compile a specific version of `controller-gen` that it uses for the rest of the build. This ensures that `controller-gen` version is consistent across all machines. I believe this setup was inherited from the CAPI provider template.

zawachte-msft: probably inherited from CAPZ. but they cleaned this up a while back... we should too.

zawachte-msft: I will clean in a future pr. Thanks for bumping up!

zawachte-msft:
If this is the route we are taking for implementation. Should we just bite the bullet and rename AzureStackHCIVirtualMachineSpec?

I thought the implementation would be to have another controller for AzureStackHCIBareMentalMachine or something like that? 

But if this the path we want let's just change the AzureStackHCIVirtualMachine to something more generic before we lock in the v1 APIS?

PR #111, opened 2021-04-02
AzureStackHciLoadBalancer scale and upgrade support

nwoodmsft:
This PR adds additional support to AzureStackHciLoadBalancer for:

* Rolling upgrade
* Scale up/down
* Conditions (LoadBalancerInfrastructureReady, LoadBalancerReplicasReady)
* AzureStackHciCluster to wait for loadbalancer infrastructure to be ready
* Phases
* Events
* Fixed logging key/value duplication

With this PR...

**kubectl get azurestackhciloadbalancer**

```
NAME                    PHASE       IP           PORT   REPLICAS   CREATED   READY   UNAVAILABLE
cluster-load-balancer   upgrading   172.2.25.1   6443   1          2         1
```

**kubectl get events**

```
16m         Normal    LoadBalancerReady                   azurestackhciloadbalancer/cluster-load-balancer                       AzureStackHCILoadBalancer cluster-load-balancer is ready
6m2s        Normal    UpgradingLoadBalancer               azurestackhciloadbalancer/cluster-load-balancer                       Upgrading AzureStackHCILoadBalancer cluster-load-balancer
6m2s        Normal    ScalingUpLoadBalancer               azurestackhciloadbalancer/cluster-load-balancer                       Scaling up AzureStackHCILoadBalancer cluster-load-balancer
3m2s        Normal    ScalingDownLoadBalancer             azurestackhciloadbalancer/cluster-load-balancer                       Scaling down AzureStackHCILoadBalancer cluster-load-balancer
```

**conditions / phase**

```
"conditions": [
	{
		"lastTransitionTime": "2021-04-02T00:18:36Z",
		"reason": "NoReplicasReady",
		"severity": "Info",
		"status": "False",
		"type": "LoadBalancerInfrastructureReady"
	},
	{
		"lastTransitionTime": "2021-04-02T00:19:27Z",
		"reason": "ScalingUp",
		"severity": "Info",
		"status": "False",
		"type": "LoadBalancerReplicasReady"
	}
],
"phase": "provisioning",
```


controllers/azurestackhcicluster_controller.go
```diff
@@ -226,12 +227,6 @@ func (r *AzureStackHCIClusterReconciler) reconcileAzureStackHCILoadBalancer(clus
 		return true, nil
 	}
 
-	// if there are some existing control plane endpoints, skip AzureStackHCILoadBalancer reconcile
```
pradipd, line 229: Will there be a way to specify no load balancers?

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -93,8 +91,9 @@ func (r *AzureStackHCILoadBalancerReconciler) Reconcile(req ctrl.Request) (_ ctr
 		return reconcile.Result{}, fmt.Errorf("Expected Cluster OwnerRef is missing from AzureStackHCILoadBalancer %s", req.Name)
 	}
 
-	azureStackHCICluster := &infrav1.AzureStackHCICluster{}
+	logger = logger.WithValues("cluster", cluster.Name)
```
pradipd, line 94: what does this do?

pradipd: or rather, why do we have to reassign logger.
Can we just do 
```
logger.WithValue()
```

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -145,113 +148,73 @@ func (r *AzureStackHCILoadBalancerReconciler) Reconcile(req ctrl.Request) (_ ctr
...
 
-	vm, err := r.reconcileNormalVirtualMachine(loadBalancerScope, clusterScope)
-	if err != nil {
-		if apierrors.IsAlreadyExists(err) {
-			clusterScope.Info("AzureStackHCIVirtualMachine already exists")
-			return reconcile.Result{}, nil
-		}
+	if lbs.AzureStackHCILoadBalancer.Spec.Replicas == nil {
+		err := errors.Errorf("No replica count was specified for loadbalancer %s", lbs.Name())
+		r.Recorder.Eventf(lbs.AzureStackHCILoadBalancer, corev1.EventTypeWarning, "InvalidConfig", err.Error())
```
pradipd, line 163: Are nil and 0 considered the same?
The mgmt cluster won't have an LB, so, I hope this won't spam the event log

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -145,113 +148,73 @@ func (r *AzureStackHCILoadBalancerReconciler) Reconcile(req ctrl.Request) (_ ctr
...
-		r.Recorder.Eventf(loadBalancerScope.AzureStackHCILoadBalancer, corev1.EventTypeWarning, "FailureReconcileLB", errors.Wrapf(err, "Failed to reconcile AzureStackHCILoadBalancer").Error())
+		r.Recorder.Eventf(lbs.AzureStackHCILoadBalancer, corev1.EventTypeWarning, "FailureReconcileLB", errors.Wrapf(err, "Failed to reconcile LoadBalancer service").Error())
+		conditions.MarkFalse(lbs.AzureStackHCILoadBalancer, infrav1.LoadBalancerInfrastructureReadyCondition, infrav1.LoadBalancerServiceReconciliationFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
 		return reconcile.Result{}, err
 	}
 
-	// wait for ip address to be exposed
-	if loadBalancerScope.Address() == "" {
-		err := r.reconcileLoadBalancerAddress(loadBalancerScope, clusterScope)
+	if lbs.Address() == "" {
```
pradipd, line 182: This seems like an odd pattern.
if address is empty
   reconileAddress()
   check if address is empty

Shouldn't all this logic belong in reconcileLBServiceStatus

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -262,14 +225,18 @@ func (r *AzureStackHCILoadBalancerReconciler) reconcileLoadBalancerAddress(loadB
...
+	if loadBalancerScope.Address() == "" {
+		loadBalancerScope.SetAddress(*((*lb.FrontendIPConfigurations)[0].IPAddress))
+	}
+
+	loadBalancerScope.SetReadyReplicas(int32(lb.ReplicationCount))
 	return nil
 }
 
-func (r *AzureStackHCILoadBalancerReconciler) reconcileLoadBalancer(loadBalancerScope *scope.LoadBalancerScope, clusterScope *scope.ClusterScope) error {
+func (r *AzureStackHCILoadBalancerReconciler) reconcileLoadBalancerService(loadBalancerScope *scope.LoadBalancerScope, clusterScope *scope.ClusterScope) error {
```
pradipd, line 239: nit: I think reconcileLoadBalancerObject is a little more consistent with our jargon.  We usually talk about the load balancer object being created.

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -287,50 +254,28 @@ func (r *AzureStackHCILoadBalancerReconciler) reconcileLoadBalancer(loadBalancer
...
-			if !apierrors.IsNotFound(err) {
-				return errors.Wrapf(err, "failed to get AzureStackHCIVirtualMachine %s", vmName)
-			}
-		}
-	}
-
-	return nil
-}
-
 func (r *AzureStackHCILoadBalancerReconciler) reconcileDeleteLoadBalancer(loadBalancerScope *scope.LoadBalancerScope, clusterScope *scope.ClusterScope) error {
```
pradipd, line 279: nit: I suppose this should be reconcileDeleteLoadBalancer*Service* to be consistent with reconcileLoadBalancer*Service*

controllers/azurestackhciloadbalancer_virtualmachine.go
```diff
@@ -0,0 +1,346 @@
...
+		if err != nil {
+			return reconcile.Result{}, errors.Wrapf(err, "failed to scale up loadbalancer")
+		}
+
+		return reconcile.Result{Requeue: true, RequeueAfter: time.Minute}, nil
+	}
+
+	// check if we need to scale down
+	if r.isScaleDownRequired(lbs) {
+		if !r.replicasAreUpgrading(lbs) {
```
pradipd, line 77: why is the scale down condition conditional on whether we are upgrading?  i.e. why not always add this condition?

controllers/azurestackhcicluster_controller.go
nwoodmsft: To properly support scale, yes you can do that. The LoadBalancerInfrastructureReady condition will show that there are no replicas available.

controllers/azurestackhciloadbalancer_controller.go
nwoodmsft: Same pattern from other controllers across CAPH. Previously this LB code was duplicating fields in the output (e.g. the LB name would appear twice).

nwoodmsft: We're basically adding a new field that we want the logger to always output (in this case, the cluster name). I believe the assignment is necessary to update the logger, at least that was my read on it.

controllers/azurestackhciloadbalancer_controller.go
nwoodmsft: There is no AzureStackHciLoadBalancer resource specified in the mgmt yaml. Cloud Operator PR removes it. All tested.

nwoodmsft: nil and zero are not the same. Zero is allowed during scaling of a k8s resource. Nil is invalid as the caller never specified a replica count for the LB.

nwoodmsft: Note though, that the CRD defaults to 1 replica...so if we see a nil it's a strange case. Just handling it in case.

controllers/azurestackhciloadbalancer_controller.go
nwoodmsft: Object feels to vague for me inside of a kubernetes controller. From CAPH's perspective, we always refer to "cloud services" and we call into services to create MOC entities. This is an existing naming pattern.

controllers/azurestackhciloadbalancer_virtualmachine.go
nwoodmsft: This was just to preserve the condition reason during upgrade (inside the "if" we are just updating a condition, not taking any other action). 

An upgrade is really just an orchestrated scale up followed by a scale down (multiple times until everything is upgraded via this rolling strategy). During the upgrade flow, we ideally want the condition reason to show that we are "upgrading" and not to flipflop between scaling up / scaling up (i'd prefer those condition reasons to only be present when we are scaling due to a requested scale operation).

Let me check though...this same if statement should probably be present on ScaleUp too. I need to check why it isn't.

controllers/azurestackhciloadbalancer_controller.go
nwoodmsft: Sure can update

controllers/azurestackhciloadbalancer_controller.go
nwoodmsft: I looked at this again. but the pattern looks the same as before and it seems to make sense to me at least :D

1) Reconcile the service (i.e. create the LB cloud resource). If it fails, log an event and change the condition to indicate that LB service reconciliation failed.

2) If the service was successfully reconciled, check to see if the LB has an address yet. If not, we reconcile status from the service (i.e. gather latest info about the LB including  addresses, replica counts etc). If that fails, we log an event and change the condition to indicate that LB service status reconciliation failed. If it succeeded, we check to see if we now have an address for the LB. If no address is present, we change the condition to indicate that the address is unavailable.

3) If the service status was reconciled and we have an address, we now look at replicas to make sure we have at least one replica servicing the LB. If no replicas are ready, we change the condition to indicate that no replicas are ready. If we have at least one replica ready then we change the condition to indicate that the LB is ready + log an event.

controllers/azurestackhciloadbalancer_controller.go
pradipd: ok. just want to make sure.

controllers/azurestackhciloadbalancer_controller.go
pradipd: ok. just curious

controllers/azurestackhciloadbalancer_controller.go
pradipd: ok.

controllers/azurestackhciloadbalancer_virtualmachine.go
pradipd: Ah, I see.  Maybe just a small comment would help the next person.

pradipd (approved):

zawachte-msft (approved):

PR #116, opened 2021-04-15
Merge baremetal caph updates into ecf branch

romoh:
Move commits from BareMetal2 into ecf branch


go.mod
```diff
@@ -23,4 +23,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.5.2
 )
 
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.4.0+incompatible
+replace (
```
romoh, line 26: @jiria - what should we do here?

jiria: We probably do not want to keep the overrides, right? Sorry for not noticing this earlier.

romoh: Good catch, will send another PR.

romoh: This was resolved before merging the PR
https://github.com/microsoft/cluster-api-provider-azurestackhci/pull/117/files

jiria: Ah, thanks.

PR #121, opened 2021-04-23
Fix retrieval logic for existing bare-metal machines.

cwize1:
Location value is being provided instead of a group value.


jiria (approved):

romoh (approved):
Good catch.

PR #123, opened 2021-04-27
Use new moc package's providerid functions.

cwize1:
Use the new `providerid.FormatProviderID` function in MOC package
instead of manually formatting the provider ID. This also inherently
adds support for bare-metal provider IDs.

### DIFF
--- a/controllers/azurestackhcimachine_controller.go
+++ b/controllers/azurestackhcimachine_controller.go
@@ -21,13 +21,12 @@ import (
 	"context"
 	"time"
 
-	"fmt"
-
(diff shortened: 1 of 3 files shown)


jiria (approved):

romoh (approved):
General comment, as part of the multiple changes do you also want to update go.mod dependencies to point to your equivalent updates/ecf latest?

cwize1:
> General comment, as part of the multiple changes do you also want to update go.mod dependencies to point to your equivalent updates/ecf latest?

Yes, as part of the commit process, I will do a cascading set of `go.mod` updates.

PR #129, opened 2021-05-20
Support for custom size vms - based on 0.3.8.

gmileka:


bfjelds (approved):

madhanrm:
Just want to call out here, we shouldnt be exposing custom vm sizes and try to define required size skus

bfjelds (approved):

PR #132, opened 2021-05-21
Support for custom size vms - based on 0.3.5

gmileka:


go.mod
```diff
@@ -24,3 +24,5 @@ require (
 )
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.4.0+incompatible
+
+replace github.com/microsoft/moc-sdk-for-go v0.10.1-alpha.1 => ../moc-sdk-for-go
```
gmileka, line 28: This is necessary since HardwordProfile.CustomSize did not exist in v0.10.1-alpha.1.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/go-autorest/autorest/to v0.3.0
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/go-logr/logr v0.1.0
-	github.com/microsoft/moc v0.10.1-alpha.4
+	github.com/microsoft/moc v0.10.9
```
gmileka, line 10: This is auto-changed by the build. It is a side effect of redirecting to latest moc-sdk-for-go.

go.mod
bfjelds: do we have a branch for moc-sdk-for-go?  if so, we can update go.mod to point at the CustomSize commit.

gmileka: I have to point go.mod to the commit of moc-sdk-for-go that has the CustomSize. I'll do that change in the next round.

bfjelds (approved):

PR #189, opened 2022-07-06
Update dockerfile to use base image from mcr

Daki-23:


PavanNeerudu:
CLosing this PR due to last inactive state being long ago and the image was moved to mariner's distroless in the last PR. Thank you.

PR #199, opened 2023-02-08
Update capi to 1.2.10

dineshba:
As per this [migration steps](https://release-1-2.cluster-api.sigs.k8s.io/developer/providers/v1.1-to-v1.2.html), below are the changes.

### Done:
1. Updated go version to 1.18.x
2. Updated controller-runtime `v0.11.x => v0.12.3`
3. Added kubernetes marker to describe the merge-strategy for server side apply.
4. Logging via component-base/logs and deprecation of klog flags

### Changes skipped (which are present in above link):
1. As there no existing immutability checks for InfrastructureMachineTemplate, skipping the validator changes.
2. No changes for cert-manager version


api/v1beta1/types.go
```diff
@@ -93,6 +93,8 @@ type VnetSpec struct {
 }
 
 // Subnets is a slice of Subnet.
+// +listType=map
```
nks5295, line 96: do the generated crd base manifests change after adding these new markers?

PR #201, opened 2023-02-18
Add warning for node out of capacity

HariPrasadV:
Added a new error type to catch when a node is out of capacity during VM creation, so that we can retry and allocate the VM on a different node if possible, without erroring out.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/go-autorest/autorest/to v0.4.0
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/go-logr/logr v1.2.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -87,6 +87,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
+replace github.com/microsoft/moc => /home/hp/repos/moc
```
HariPrasadV, line 90: Todo: Change this after merge the new error type and releasing moc

pradipd: @HariPrasadV  Can we update this now that we have the package?

pradipd (approved):

apluchik (approved):

PR #203, opened 2023-02-27
Moc group creation

shyamlirao:
Currently when a cluster is created using akshci cli, we create MocGroup in kva. This is fine since the cluster name is known beforehand and we can create moc group as clustergroup-<clustername>. 
When az cli is used for cluster creation, actual cluster name given in cli is different from the one created. This because we add a random number (azure UUID) in the end, for example: clustername-<azure id>. So, in this case we cannot create the moc group beforehand. We want moc group to be created by lower layers therefore we are introducing it in CAPI layer.

### DIFF
--- a/cloud/services/groups/groups.go
+++ b/cloud/services/groups/groups.go
@@ -26,6 +26,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	TagKeyClusterGroup = "createdBy"
+	TagValClusterGroup = "cloud-operator"
+)
+
 // Spec specification for group
 type Spec struct {
 	Name     string
@@ -49,6 +54,7 @@ func (s *Service) Get(ctx context.Context, spec interface{}) (interface{}, error
 
 // Reconcile gets/creates/updates a group.
 func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
+
 	groupSpec, ok := spec.(*Spec)
 	if !ok {
 		return errors.New("Invalid group specification")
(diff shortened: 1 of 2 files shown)


cloud/services/groups/groups.go
```diff
@@ -58,37 +63,55 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		// group already exists, cannot update since its immutable
 		return nil
 	}
+	//adding tag to group
```
raghavendra-nataraj, line 66: We may want to check if tags already exists before adding new ones, so that we don't overwrite existing ones

raghavendra-nataraj (approved):

cloud/services/groups/groups.go
```diff
@@ -26,6 +26,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	TagKeyClusterGroup = "createdBy"
+	TagValClusterGroup = "cloud-operator"
```
dineshba, line 31: Can we have the value as "caph" instead of cloud-operator?

If group is created by caph, caph will delete it during cluster reconcile. 

Thoughts?

shyamlirao: sure, we can update it.

cloud/services/groups/groups.go
```diff
@@ -58,37 +63,55 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		// group already exists, cannot update since its immutable
 		return nil
 	}
+	//adding tag to group
+	tag := make(map[string]*string, 1)
+	cloudop := TagValClusterGroup
```
zawachte, line 68: why are we leaking all this cloudoperator context into caph?

cloud/services/groups/groups.go
```diff
@@ -26,6 +26,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	TagKeyClusterGroup = "createdBy"
```
zawachte, line 30: should this tag be called something else? Looks like it is set in the case where caph "adopts" the resourceGroup but doesn't create it. In that case its weird to have a tag named "createdBy" right?

cloud/services/groups/groups.go
```diff
@@ -58,37 +63,55 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
+	} else if err != nil {
+		return err
 	}
-	if err != nil {
-		return errors.Wrapf(err, "failed to delete group %s", groupSpec.Name)
+	groupObj := (*group)[0]
+	value, ok := groupObj.Tags[TagKeyClusterGroup]
+	// delete only if created by cloud operator
+	if ok && (value != nil && *value == TagValClusterGroup) {
+		err := s.Client.Delete(ctx, groupSpec.Location, groupSpec.Name)
```
zawachte, line 104: Is it safe to delete this group here? What if there is a AksHciNetwork associated with it? That is all handled at the higher level right?

@pradipd thoughts?

cloud/services/groups/groups.go
zawachte: this is a flow for caph to adopt a resource group right? nothing really to do with cloudoperator.

cloud/services/groups/groups.go
raghavendra-nataraj: Caph is not adopting the group. If the group does not exist when the cluster is created , CAPH and created the group. This is to know if the caph can delete the group it created during cluster deletion. This is to avoid the behavior we have in AksHci where KVA creates the group and HybridAks where group would be created CAPH.  We would ideally want to move the group creation to CAPH for AksHci too

cloud/services/groups/groups.go
raghavendra-nataraj: The group is deleted only if the caph is creating it during cluster creation. So it should not delete the group that was created by higher levels.

cloud/services/groups/groups.go
zawachte: Thanks for clearing this up. I would prefer "ownedBy" but that is really a nit.

cloud/services/groups/groups.go
zawachte: Spoke to ragha offline. cleared it up.

This tag associated and variables should be mentioned caph not cloudop.

cloud/services/groups/groups.go
```diff
@@ -58,37 +63,55 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
...
 	if err != nil {
 		return err
 	}
-
 	klog.V(2).Infof("successfully created group %s", groupSpec.Name)
 	return err
 }
 
-// Delete deletes a group.
+// Delete deletes a group if group is created by cloud operator
```
zawachte, line 85: owned by caph/self.

cloud/services/groups/groups.go
shyamlirao: yes, will make the change.

raghavendra-nataraj (dismissed):

zawachte (approved):

controllers/azurestackhcicluster_reconciler.go
```diff
@@ -92,6 +106,14 @@ func (r *azureStackHCIClusterReconciler) Delete() error {
 			return errors.Wrapf(err, "failed to delete keyvault %s for cluster %s", r.scope.Name(), r.scope.Name())
 		}
 	}
+	groupSpec := &groups.Spec{
+		Name:     r.scope.GetResourceGroup(),
+		Location: r.scope.Location(),
+	}
+
+	if err := r.groupSvc.Delete(r.scope.Context, groupSpec); err != nil {
```
zawachte, line 114: Since we are hiding all the logic about when and when not to delete a group in the cloud service layer ... do we want to add a comment here for future readers to have a hint about this group deletion complication?

shyamlirao: added comment

nwoodmsft (approved):

PR #206, opened 2023-03-17
Replace math/rand with crypto/rand

ganeshamohana:


pkg/util/util.go
```diff
@@ -42,7 +38,12 @@ func GetAzureStackHCIMachinesInCluster(ctx context.Context, controllerClient cli
 func RandomAlphaNumericString(n int) string {
 	result := make([]byte, n)
 	for i := range result {
-		result[i] = charSet[rnd.Intn(len(charSet))]
+		val, err := rand.Int(rand.Reader, big.NewInt(int64(len(charSet))))
+		if err != nil {
+			result[i] = charSet[i]
+		} else {
```
raghavendra-nataraj, line 44: Can we think of an alternative here. We are not sure when we would hit this

raghavendra-nataraj (changes requested):

pkg/util/util.go
ganeshamohana: We can propagate the error back to the caller and let them handle it. I can't think of any other ways to handle when we can't generate a random number.

raghavendra-nataraj (dismissed):

pkg/util/util.go
```diff
@@ -2,20 +2,11 @@ package util
...
-	"math/rand"
-	"time"
 
 	infrav1 "github.com/microsoft/cluster-api-provider-azurestackhci/api/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-)
-
-const (
-	charSet = "abcdefghijklmnopqrstuvwxyz0123456789"
```
nwoodmsft, line 14: Does this charset limitation/guarantee continue to hold with the switch to moc-pkg rand implementation?

pkg/util/util.go
```diff
@@ -39,10 +30,6 @@ func GetAzureStackHCIMachinesInCluster(ctx context.Context, controllerClient cli
...
 
 // RandomAlphaNumericString returns a random alphanumeric string.
-func RandomAlphaNumericString(n int) string {
-	result := make([]byte, n)
-	for i := range result {
-		result[i] = charSet[rnd.Intn(len(charSet))]
-	}
-	return string(result)
+func RandomAlphaNumericString(n int) (string, error) {
+	return rand.RandomString(n)
```
nwoodmsft, line 34: If we are reducing this code down to just a wrapper around calling rand.RandomString(n) then it doesn't make a lot of sense to me to keep this RandomAlphaNumericString function. I would remove this function and/or file and just call rand.RandomString(n) directly wherever needed.

pkg/util/util.go
ganeshamohana: Yes it'll hold, although moc-pkg might not use the exact same charset for random string generation, it'll guarantee the randomness, alphanumeric and no bad words properties of the string.

pkg/util/util.go
ganeshamohana: Thank you for the suggestion, I was on the fence to remove this function but now I'll go ahead and remove this util function.

go.mod
```diff
@@ -49,13 +50,13 @@ require (
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/magiconair/properties v1.8.7 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 // indirect
-	github.com/microsoft/wmi v0.5.0-alpha.11 // indirect
+	github.com/microsoft/wmi v0.5.0-alpha.13 // indirect
```
madhanrm, line 53: This repo shouldnt have dependency with wmi.
can you also bump up github.com/microsoft/moc to remove this dependency.

ganeshamohana: I think it's because moc-pkg has wmi as dependency and since we're using moc-pkg here, wmi is added as indirect dependency.

nwoodmsft (approved):

PR #208, opened 2023-03-27
Ensure imageName is not empty for custom image

apluchik:
Making the getVMImage logic consistent with 
https://github.com/microsoft/cluster-api-provider-azurestackhci/blob/922043bea4532118bb0322e015ee653aad1b0209/controllers/azurestackhcimachine_controller.go#L414


apluchik:
@nwoodmsft @pradipd could you please take a look when you have a chance?

nwoodmsft (approved):

PR #213, opened 2023-05-26
Add moc operation log

mogliang:
As part of telemetry enhancement https://msazure.visualstudio.com/msk8s/_workitems/edit/23541342
Adding logs to record create/update/delete requests from caph to moc.

### DIFF
--- a/cloud/logutils.go
+++ b/cloud/logutils.go
@@ -1,0 +1,75 @@
+package azurestackhci
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	mocerrors "github.com/microsoft/moc/pkg/errors"
+	"k8s.io/klog"
+)
+
+type MocResourceType string
+
+const (
+	LoadBalancer     MocResourceType = "LoadBalancer"
+	VipPool          MocResourceType = "VipPool"
+	VirtualNetwork   MocResourceType = "VirtualNetwork"
+	NetworkInterface MocResourceType = "NetworkInterface"
+	Disk             MocResourceType = "Disk"
+	VirtualMachine   MocResourceType = "VirtualMachine"
+	KeyVault         MocResourceType = "KeyVault"
+	Secret           MocResourceType = "Secret"
+	Group            MocResourceType = "Group"
+)
+
+type MocOperation string
+
+const (
+	CreateOrUpdate MocOperation = "CreateOrUpdate"
+	Delete         MocOperation = "Delete"
+	Get            MocOperation = "Get"
+)
+
+type OperationLog struct {
+	Timestamp      string      `json:"timestamp"`
+	ParentResource string      `json:"parent_resource"`
+	Resource       string      `json:"resource"`
+	FilterKeyword  string      `json:"filter_keyword"`
+	Action         string      `json:"action"`
+	Params         interface{} `json:"params"`
+	ErrorCode      string      `json:"error_code"`
+	Message        string      `json:"msg"`
+}
+
+func WriteMocOperationLog(operation MocOperation, crResourceName string, mocResourceType MocResourceType, mocResourceName string, params interface{}, err error) {
+	errcode := "0"
+	message := ""
+	if err != nil {
+		errcode = mocerrors.GetErrorCode(err)
+		message = err.Error()
+	}
+
+	oplog := OperationLog{
+		Timestamp:      time.Now().Format(time.RFC3339),
+		ParentResource: crResourceName,
+		Resource:       fmt.Sprintf("%s/%s", mocResourceType, mocResourceName),
+		FilterKeyword:  "RESOURCE_ACTION",
+		Action:         string(operation),
+		Params:         params,
+		ErrorCode:      errcode,
+		Message:        message,
+	}
+
+	jsonData, err := json.Marshal(oplog)
+	if err != nil {
+		klog.Error("Unable to serialize operation log object. ", crResourceName)
+	} else {
+		klog.Info(string(jsonData))
+	}
+}
+
+func GenerateMocResourceName(nameSegments ...string) string {
+	return strings.Join(nameSegments, "/")
+}
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -19,6 +19,7 @@ package scope
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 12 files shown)


cloud/logutils.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		errcode = mocerrors.GetErrorCode(err)
+		message = err.Error()
+	}
+
+	oplog := OperationLog{
+		Timestamp:      time.Now().Format(time.RFC3339),
+		ParentResource: crResourceName,
+		Resource:       fmt.Sprintf("%s/%s", mocResourceType, mocResourceName),
+		FilterKeyword:  "RESOURCE_ACTION",
```
pradipd, line 62: What does this mean?

cloud/logutils.go
```diff
@@ -0,0 +1,79 @@
...
+	message := ""
+	if err != nil {
+		errcode = mocerrors.GetErrorCode(err)
+		message = err.Error()
+	}
+
+	oplog := OperationLog{
+		Timestamp:      time.Now().Format(time.RFC3339),
+		ParentResource: crResourceName,
+		Resource:       fmt.Sprintf("%s/%s", mocResourceType, mocResourceName),
```
pradipd, line 61: Why combine resource type and name into a single string?  why not have 2 seperate strings instead of concatenating?

cloud/logutils.go
```diff
@@ -0,0 +1,79 @@
...
+	FilterKeyword  string      `json:"filter_keyword"`
+	Action         string      `json:"action"`
+	Params         interface{} `json:"params"`
+	ErrorCode      string      `json:"error_code"`
+	Message        string      `json:"msg"`
+}
+
+func WriteMocOperationLog(operation MocOperation, crResourceName string, mocResourceType MocResourceType, mocResourceName string, params interface{}, err error) {
+	if (operation == Delete || operation == Get) && ResourceNotFound(err) {
+		err = nil
```
pradipd, line 48: Are we sure we want to swallow this error?  
What if the error/bug is that we returned resource not found when we didn't expect it.
E.g. let's say MOC stack has a *bug* and createorupdate() returned resource not found.  Our telemetry would not reflect that error.
If we want our teleemtry to be effective, we should follow the calls inputs and outputs as is.
Any interpertation should happen in the queries.

cloud/scope/cluster.go
```diff
@@ -116,6 +117,11 @@ func (s *ClusterScope) GetAuthorizer() auth.Authorizer {
 	return s.Authorizer
 }
 
+// GetCustomResourceTypeWithName return the type and name string of the associated CR.
+func (s *ClusterScope) GetCustomResourceTypeWithName() string {
+	return fmt.Sprintf("%s/%s/%s", s.Kind(), s.Namespace(), s.Name())
```
pradipd, line 122: similar to above question.  why combine into a string?
this is going to a dependency on the code and the queries.  
if some developer/pm/manager wants to write an ad-hoc query, they will have to know that the format is kind/namespace/name.  And they will have to either ask around or try to find this code.

cloud/services/disks/disks.go
```diff
@@ -39,6 +39,8 @@ func (s *Service) Get(ctx context.Context, spec interface{}) (interface{}, error
 		return storage.VirtualHardDisk{}, errors.New("Invalid Disk Specification")
 	}
 	disk, err := s.Client.Get(ctx, s.Scope.GetResourceGroup(), "", diskSpec.Name)
+	azurestackhci.WriteMocOperationLog(azurestackhci.Get, s.Scope.GetCustomResourceTypeWithName(), azurestackhci.Disk,
+		azurestackhci.GenerateMocResourceName(s.Scope.GetResourceGroup(), diskSpec.Name), nil, err)
```
pradipd, line 43: Should we log the returned object as well?
it would help trace the call through layers.  (e.g. moc stack returned object x. and in CAPH we see object Y -- that could be a potential bug).

pradipd:
Thanks!!! I appreciate you starting on this. 
Do you have an example of what the telemetry data would look like?
And a sample query?

cloud/logutils.go
```diff
@@ -0,0 +1,79 @@
...
+type MocOperation string
+
+const (
+	CreateOrUpdate MocOperation = "CreateOrUpdate"
+	Delete         MocOperation = "Delete"
+	Get            MocOperation = "Get"
+)
+
+type OperationLog struct {
+	Timestamp      string      `json:"timestamp"`
```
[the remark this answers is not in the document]
chenditc: Taking the ideal of OpenTelemetry. OpenTelemetry correlate event by span id. Each event will spit out parent span id and its span id. This create a tree structure events.

We will correlate across component by parent resource id and its own resource id. We will scope the events to a single management cluster by deployment ID.

cloud/scope/cluster.go
chenditc: The idea is to have one thing which works like span id across layers. We don't expect people directly query this log, they will come from build in query and have some example to start with.

cloud/services/disks/disks.go
chenditc: Currently we don't log the object detail in or out, like crd spec and MOC config. We do want to add them.

But we are worry about:
1. The size of the object might be very big and exceed the limit of kusto cell.
2. The volume will be very big and not allowed to go through the geneva warm path.
3. The object detail might contains PII.

So for now, we don't add full object detail. In the future, we might add a new event like "RESOURCE_DETAIL" to log the full object in and out.

cloud/logutils.go
mogliang: thanks for suggestion. agree that better to return original error, and filter in kusto

mogliang:
> Thanks!!! I appreciate you starting on this. Do you have an example of what the telemetry data would look like? And a sample query?

MicrosoftTeams-image

We haven't upload to kusto yet, chendi will work on that later.

cloud/logutils.go
pradipd: In the query, how do you know if you are dealing with a moc object (and hence moc resource type/moc resource name) vs k8s CRD vs a Azure resource?

A doc explaining how this comes together would be useful here.

cloud/services/disks/disks.go
pradipd: ok

cloud/scope/cluster.go
pradipd: ok. again a doc would really help here.

pradipd (approved):

cloud/logutils.go
pradipd: How do you know there won't be any name collisions in the various layers?
E.g. for moc, there is cloudagent and nodeagent.  they would both container a moc resource type and moc resource name.  And they may (or may not) be the same.  I can't recall if the name is different.

chenditc: The idea is not to worry if the resource is a moc resource or k8s resource or Azure resource. Like in OpenTelemetry, we won't worry about where the correlation id come from. Just take the ID and get related events. 

For name collision, the assumption is: If upper layer only have resource type and resource name to identify a moc resource. Then from the upper layer's perspective, they are same, we should put related events together.

cloud/scope/cluster.go
chenditc: Copied doc to the well known place: https://github.com/microsoft/AKS-Edge-Docs/blob/main/ARC-HCI/CorrelationThroughLayers.md

PR #216, opened 2023-06-07
Fix go install.sh

chenditc:


.pipelines/build.yaml
```diff
@@ -40,6 +40,8 @@ jobs:
           version: "1.19.9"
 
       - script: |
+          set -e
```
pradipd, line 43: The build is failing because of the set -e and the mkdir out/ command.
i would change the mkdir command to "[ -d out] || mkdir out" to fix the build.

PR #221, opened 2023-07-11
get and log moc deployment id

lpdnju:


cloud/logutils.go
```diff
@@ -73,3 +77,29 @@ func WriteMocOperationLog(operation MocOperation, crResourceName string, mocReso
...
+func WriteMocDeploymentIdLog(ctx context.Context, cloudAgentFqdn string, authorizer auth.Authorizer) {
+	deploymentId, err := getHealthClient(cloudAgentFqdn, authorizer).GetDeploymentId(ctx)
+	if err != nil {
+		klog.Error("Unable to get moc deployment id. ", err)
+	} else {
+		klog.Infof("MOC Deployment Id: %s", deploymentId)
+	}
+}
+
+func getHealthClient(cloudAgentFqdn string, authorizer auth.Authorizer) *health.HealthClient {
```
zawachte, line 92: Did we consider treating the health client like other moc services and the client pattern we have for them? Is there a need to special case healthClient?

lpdnju: I think other service&client is more used by controller for CRUD of moc resources. HealthClient is more like basic or internal service depended by other service and currently it will only be used for logging.

lpdnju: > Did we consider treating the health client like other moc services and the client pattern we have for them? Is there a need to special case healthClient?

I think other service&client is more used by controller for CRUD of moc resources. HealthClient is more like basic or internal service depended by other service and currently it will only be used for logging.

zawachte: I still think we should follow the same pattern for it incase there are more "internal services" in future and it becomes unruly to manage in util files.

lpdnju: > I still think we should follow the same pattern for it incase there are more "internal services" in future and it becomes unruly to manage in util files.

as you know now all services import logutil. I have concern if all other services import util and util import one service, it will be confusing about circular dependency.

zawachte: ok. I see that moc-operator is also breaking its pattern for the health client so its ok by me. https://dev.azure.com/msazure/msk8s/_git/moc-operator/commit/527027733b74bc2892c4ebb308c2f6a1489592fe?_a=compare&path=%2Fpkg%2Ferrors%2Ferrors.go 

w

zawachte:
LGTM. just worried about the version compatibility issues we spoke about offline.

cloud/logutils.go
pradipd: why can't you move logutils into its own log package?  that would break the dependency and keep the pattern.

pradipd: Looking at the code, I think the telemetry code was just thrown into azurestackhci package.  
I do not feel like that is the correct location.  I think we should make a telemetry (or log) package). 
I *think* that would break the dependency and be a cleaner separation of packages.

pradipd (changes requested):
Can we move the logutils to a separate package?

cloud/telemetry/logutils.go
```diff
@@ -73,3 +77,29 @@ func WriteMocOperationLog(operation MocOperation, crResourceName string, mocReso
...
+func WriteMocDeploymentIdLog(ctx context.Context, cloudAgentFqdn string, authorizer auth.Authorizer) {
+	deploymentId, err := getHealthClient(cloudAgentFqdn, authorizer).GetDeploymentId(ctx)
+	if err != nil {
+		klog.Error("Unable to get moc deployment id. ", err)
+	} else {
+		klog.Infof("MOC Deployment Id: %s", deploymentId)
+	}
+}
+
+func getHealthClient(cloudAgentFqdn string, authorizer auth.Authorizer) *health.HealthClient {
```
pradipd, line 92: Now that this is in it's own package, can't we move healthclient to services like all the other calls to moc agent?
My point was that if we move this to its own package, we would be able to break the circular dependency.

pradipd (approved):

pradipd:
Thanks!

zawachte:
Think we need to hold on this PR until we come to a conclusion about the moc versioning promises.

pradipd (changes requested):
Need to close on to handle new moc API's in the email thread.

lpdnju:
> Need to close on to handle new moc API's in the email thread.

from mail thread: we will take temporary workaround in cloud-op to unblock PR and code in CAPH is safe even if it failes to get deloymentid.

pradipd (approved):

PR #222, opened 2023-07-11
Fix createOrUpdate virtualMachine log to long issue

mogliang:
- change parent resource to cluster https://msazure.visualstudio.com/msk8s/_workitems/edit/24476850
- fix createOrUpdate virtualMachine log to long issue
- remove 'GET' mocoperation log

### DIFF
--- a/cloud/logutils.go
+++ b/cloud/logutils.go
@@ -51,6 +51,11 @@ func WriteMocOperationLog(operation MocOperation, crResourceName string, mocReso
 		message = err.Error()
 	}
 
+	// don't collect get moc operation log to minimize log size.
+	if operation == Get {
+		return
+	}
+
 	oplog := OperationLog{
 		Timestamp:      time.Now().Format(time.RFC3339),
 		ParentResource: crResourceName,
--- a/cloud/logutils.go
+++ b/cloud/logutils.go
@@ -51,11 +51,6 @@ func WriteMocOperationLog(operation MocOperation, crResourceName string, mocReso
 		message = err.Error()
 	}
 
-	// don't collect get moc operation log to minimize log size.
-	if operation == Get {
-		return
-	}
-
 	oplog := OperationLog{
 		Timestamp:      time.Now().Format(time.RFC3339),
 		ParentResource: crResourceName,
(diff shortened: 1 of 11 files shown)


chenditc (approved):

zawachte (approved):

cloud/logutils.go
```diff
@@ -51,6 +51,11 @@ func WriteMocOperationLog(operation MocOperation, crResourceName string, mocReso
 		message = err.Error()
 	}
 
+	// don't collect get moc operation log to minimize log size.
+	if operation == Get {
```
pradipd, line 55: Remove the call to WriteMocOperationLog() in the calling Get() functions.
It seems silly to call this function just to ignore Get operations.

cloud/services/virtualmachines/virtualmachines.go
```diff
@@ -186,7 +186,7 @@ func (s *Service) Reconcile(ctx context.Context, spec interface{}) error {
 		vmSpec.Name,
 		&virtualMachine)
 	azurestackhci.WriteMocOperationLog(azurestackhci.CreateOrUpdate, s.Scope.GetCustomResourceTypeWithName(), azurestackhci.VirtualMachine,
-		azurestackhci.GenerateMocResourceName(s.Scope.GetResourceGroup(), vmSpec.Name), &virtualMachine, err)
+		azurestackhci.GenerateMocResourceName(s.Scope.GetResourceGroup(), vmSpec.Name), nil, err)
```
pradipd, line 189: Note if, at some point in time, we find we need the virtualmachine object, we could just skip the customdata field.  that is where the length is coming from.

I.e. 
type Spec struct {
	Name       string
	NICName    string
	SSHKeyData []string  <-- we should probably not log this.
	Size       string
	Zone       string
	Image      infrav1.Image
	OSDisk     infrav1.OSDisk
	CustomData string <-- this is going to super long, if we skip this the virtual machine object might be reasonable size.
	VMType     compute.VMType
}

pradipd: But, for now, it is fine assuming we don't need the vm object.

cloud/logutils.go
mogliang: it's not very sure that we don't want GET operation log, maybe regret sometime later :)
Therefore i make this change so that we may need little change to revert

cloud/services/virtualmachines/virtualmachines.go
mogliang: sure, thanks for the advice. We can add vm back when we find it's useful.

cloud/logutils.go
mogliang: yeah, let me clean it up. it's not that much

pradipd (approved):

cloud/logutils.go
pradipd: Yeah, it will be easy to add back if/when we need it.
And most likely we will only want to add it back in certain places.

PR #228, opened 2023-08-14
logging moc version

lpdnju:


cloud/telemetry/logutils.go
```diff
@@ -77,11 +78,39 @@ func GenerateMocResourceName(nameSegments ...string) string {
...
+	versionPair, err := getVersionsService(scope).Get(ctx)
+	if err != nil {
+		klog.Error("Unable to get moc version. ", err)
+	} else {
+		wssdCloudAgentVersion = versionPair.WssdCloudAgentVersion
+		mocVersion = versionPair.MocVersion
+	}
+
+	infoLog := MocInfoLog{
+		MocDeploymentId:       deploymentId,
```
pradipd, line 104: Do we really need to write this multiple times?  We are writing this out in every call to moc, but, the information will be the same.  
Can we just write this out once?  Maybe when we start CAPH?

lpdnju: We have considered logging only CAPH starts,  it will increase the cost of searching because massive logs of all stamps are stored in kusto/geneva and our target is only one line logged months ago(depend on pod restart). 
Another reason is if moc info is not logged  in one api trace, we has to run another query to get moc id for diagnostic.

pradipd: ok. i'll approve.
Personally, i do not think that is a good design tradeoff.  
We are consuming the customers bandwidth and sending redundant information.  Versus having a more complex query in a microsoft cloud environment where we have tons of money and resources.
But, I also see that the redudant information is small.  So, I'm a bit torn on the tradeoff.

PR #229, opened 2023-08-31
Extend logger to include unique id per reconcile run

yepyao:


lpdnju (approved):

zawachte (dismissed):

controllers/azurestackhcivirtualmachine_controller.go
```diff
@@ -62,9 +63,9 @@ func (r *AzureStackHCIVirtualMachineReconciler) SetupWithManager(mgr ctrl.Manage
 
 // Reconcile reacts to some event on the kubernetes object that the controller has registered to handle
 func (r *AzureStackHCIVirtualMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
-	logger := r.Log.WithValues("namespace", req.Namespace, "azureStackHCIVirtualMachine", req.Name)
-
-	logger.Info("attempt reconcile resource", "name", req.NamespacedName)
+	logger := r.Log.WithValues("namespace", req.Namespace, "azureStackHCIVirtualMachine", req.NamespacedName)
```
pradipd, line 66: Nit: Is it possible to do something like:
logger := r.Log.WithValues("namespace", req.Namespace, "azureStackHCIVirtualMachine", req.NamespacedName, "reconcileId", infrav1util.GetReconcileID(ctx))

This way it is just 1 line to get the logger instead of 2 :
logger := r.Log.WithValues("azureStackHCIMachine", req.NamespacedName)
logger = infrav1util.AttachReconcileIDToLogger(ctx, logger)

My concern is that someone may forgot to add the 2nd line in the future and not realize it.

pradipd: @yepyao Changes looks ok.  so, i signed off.
But, if possible, can you consider making the above change?  i think it would make it cleaner.  
However, if you think it is too much work or not worth it, then it is ok as is.

pradipd (approved):

pradipd:
Thanks!

PR #230, opened 2023-09-08
Move AlreadyExists Catch from CreateOrUpdate to inside reconcileVirtualMachineNormal and continue

zawachte:
Today in the code, we check the error returned by reconcileVirtualMachineNormal and end the reconcile loop if it is AlreadyExists.

This is incorrect behavior and forces us to wait for periodic reconcile to finish processing the machine object. To fix this we are moving the AlreadyExists error check into reconcileVirtualMachineNormal and instead of ignoring the error and ending the reconcile, we are ignoring and continuing.

There is also of other context regarding the background of when CreateOrUpgrade can return AlreadyExists that we are still looking into and will iterate on in a future PR.

### DIFF
--- a/controllers/azurestackhcimachine_controller.go
+++ b/controllers/azurestackhcimachine_controller.go
@@ -321,7 +321,9 @@ func (r *AzureStackHCIMachineReconciler) reconcileVirtualMachineNormal(machineSc
 		//
 		// We believe this is ok as this cache issue is only going to be seen very close to when the object was
 		// initially created. We are also looking to improve this behavior by introducing live clients or polling.
-		if !apierrors.IsAlreadyExists(err) {
+		if apierrors.IsAlreadyExists(err) {
+			machineScope.Info("CreateOrUpdate in reconcileVirtualMachineNormal returned AlreadyExists", "vmName", vm.Name)
+		} else {
 			return nil, errors.Wrapf(err, "failed to CreateOrUpdate AzureStackHCIVirtualMachine %s", vm.Name)
 		}
 	}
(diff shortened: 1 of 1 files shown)


raghavendra-nataraj (approved):

controllers/azurestackhcimachine_controller.go
```diff
@@ -312,7 +308,22 @@ func (r *AzureStackHCIMachineReconciler) reconcileVirtualMachineNormal(machineSc
...
+		// 2. CreateOrUpdate is called a second time and returns AlreadyExists due to the cache reasoning.
+		//
+		// Since we are ignoring AlreadyExists, if the second call to CreateOrUpdate has updates to the object,
+		// they will not be applied silently.
+		//
+		// We believe this is ok as this cache issue is only going to be seen very close to when the object was
+		// initially created. We are also looking to improve this behavior by introducing live clients or polling.
+		if !apierrors.IsAlreadyExists(err) {
+			return nil, errors.Wrapf(err, "failed to CreateOrUpdate AzureStackHCIVirtualMachine %s", vm.Name)
+		}
```
pradipd, line 326: Can we *easily* add a telemetry point to see how frequently we see "AlreadyExists" returned?
If not, perhaps we should at least add a log message in case other things happen.

zawachte: good catch. Yes I mean log/telemetry/event ... its all the same as long as you can create a kusto query around it. Will add a log line.

zawachte: @pradipd done.

madhanrm (approved):

PR #231, opened 2023-09-13
logging correlation id

lpdnju:


api/v1beta1/azurestackhcimachine_types.go
```diff
@@ -136,6 +136,22 @@ func (c *AzureStackHCIMachine) SetConditions(conditions clusterv1.Conditions) {
 	c.Status.Conditions = conditions
 }
 
+func (c *AzureStackHCIMachine) GetCorrelationId() string {
```
pradipd, line 139: Is there a way we can minimize the cut and paste?
I know it is small, but ,we cut and pasted it 4 times.

pradipd (approved):

pradipd:
I have approved.  But, it would be nice to see if we can minimize the cut and paste.
If it is not possible, then go ahead and submit.

controllers/azurestackhcicluster_controller.go
```diff
@@ -77,6 +77,8 @@ func (r *AzureStackHCIClusterReconciler) Reconcile(ctx context.Context, req ctrl
 		return reconcile.Result{}, err
 	}
 
+	log = log.WithValues("operationId", azureStackHCICluster.GetOperationId(), "correlationId", azureStackHCICluster.GetCorrelationId())
```
pradipd, line 80: I think you can just do azureStackHCICluster.Annotations [<key>] and get the same result as your function.

pradipd (approved):

pradipd:
Thanks for updating.

PR #232, opened 2023-09-22
pass correlation id to apis for moc

lpdnju:
adding correlation id in ctx that used to call moc api.
related changes in moc: https://github.com/microsoft/moc/pull/203/files


cloud/scope/cluster.go
```diff
@@ -71,14 +72,16 @@ func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
...
+	scopeContext := diagnostics.NewContextWithCorrelationId(context.Background(), params.AzureStackHCICluster.GetAnnotations()[infrav1.AzureCorrelationIDAnnotationKey])
 	scope := &ClusterScope{
 		Logger:               *params.Logger,
 		Client:               params.Client,
 		AzureStackHCIClients: params.AzureStackHCIClients,
 		Cluster:              params.Cluster,
 		AzureStackHCICluster: params.AzureStackHCICluster,
 		patchHelper:          helper,
-		Context:              context.Background(),
+		Context:              scopeContext,
```
pradipd, line 84: So how does the correlation id get sent to moc?
i see that this is adding a new context w/ correclation id to the scope, but, I'm not understanding how moc will use it.

pradipd: is there a corresponding change in wssdcloudagent/wssdagent ?

pradipd (approved):

pradipd:
I have approved. Semantically, the changes look fine.   And we can start getting this into cloud-op.

But, I'm still don't understand how moc stack will consume this.  If there are relevant moc PR's please let me know.

cloud/scope/cluster.go
lpdnju: Yes, changes:
https://github.com/microsoft/moc-pkg/pull/420/files
https://github.com/microsoft/wssdcloudagent/pull/802
https://github.com/microsoft/wssdagent/pull/797

pradipd: ah. now i see.  Thanks!

PR #233, opened 2023-10-02
Support Non-Default StorageContainer

nwoodmsft:
Support for passing in a storage container name for the CAPH VMs (k8s and lb) to use. The storage container name will be passed to MOC SDK when we call to create a virtual machine, which should ensure that the VM is created on the requested MOC storage container. If no storage container is specified then MOC will choose a default container (the underlying default in MOC seems to be the default "MocStorageContainer" container).

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhciclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhciclusters.yaml
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
-  creationTimestamp: null
+    controller-gen.kubebuilder.io/version: v0.11.4
   name: azurestackhciclusters.infrastructure.cluster.x-k8s.io
 spec:
   group: infrastructure.cluster.x-k8s.io
@@ -688,6 +687,8 @@ spec:
                     type: integer
                   sshPublicKey:
                     type: string
+                  storageContainer:
+                    type: string
                   vmSize:
                     type: string
                 required:
@@ -915,9 +916,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhcimachinetemplates.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhcimachinetemplates.yaml
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
-  creationTimestamp: null
+    controller-gen.kubebuilder.io/version: v0.11.4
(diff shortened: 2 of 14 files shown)


config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhciclusters.yaml
```diff
@@ -688,11 +687,14 @@ spec:
...
                   sshPublicKey:
                     type: string
+                  storageContainer:
+                    type: string
                   vmSize:
                     type: string
                 required:
                 - image
                 - sshPublicKey
+                - storageContainer
```
zawachte, line 697: I don't think we can make this required

config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhcimachinetemplates.yaml
```diff
@@ -99,33 +98,6 @@ spec:
                         type: object
                       location:
                         type: string
```
zawachte, line 100: I am a bit confused by this removal.

nijosmsft: yes, we need this sdncni support. can we please add it back? same for azurestackhcivirtualmachines. thanks.

nwoodmsft: This was auto-generated, not manually removed, so does this mean there is some issue with the current code/structure that is causing this to be removed?

config/crd/bases/infrastructure.cluster.x-k8s.io_azurestackhciclusters.yaml
nwoodmsft: Agree, marked as optional

zawachte (approved):

controllers/azurestackhcivirtualmachine_reconciler.go
```diff
@@ -230,15 +230,16 @@ func (s *azureStackHCIVirtualMachineService) createVirtualMachine(nicName string
...
+			Name:             s.vmScope.Name(),
+			NICName:          nicName,
+			SSHKeyData:       decodedKeys,
+			Size:             s.vmScope.AzureStackHCIVirtualMachine.Spec.VMSize,
+			OSDisk:           s.vmScope.AzureStackHCIVirtualMachine.Spec.OSDisk,
+			Image:            s.vmScope.AzureStackHCIVirtualMachine.Spec.Image,
+			CustomData:       *s.vmScope.AzureStackHCIVirtualMachine.Spec.BootstrapData,
+			Zone:             vmZone,
+			VMType:           vmType,
+			StorageContainer: s.vmScope.StorageContainer(),
```
raghavendra-nataraj, line 242: Should we have a validation to check if the storage container exists in Moc?

nwoodmsft: Technically that check will happen today as the validation of container existence occurs in KVA as part of Validate() API, but I agree we can work on adding this kind of check to harden the lower layers such as CAPH also. Given the coverage is already present in KVA Validate, can we fast-follow after this PR with that extra validation?

raghavendra-nataraj (approved):

controllers/azurestackhcivirtualmachine_reconciler.go
raghavendra-nataraj: Sure. Thank you

PR #238, opened 2023-10-26
update correlation id of virtual machine CRs when create/delete loadbalancer

lpdnju:


controllers/azurestackhciloadbalancer_virtualmachine.go
```diff
@@ -130,6 +131,13 @@ func (r *AzureStackHCILoadBalancerReconciler) reconcileDeleteVirtualMachines(loa
 
 	for _, vm := range vmList {
 		if vm.GetDeletionTimestamp().IsZero() {
+			// update correlationId before deletion
+			infrav1util.CopyCorrelationId(loadBalancerScope.AzureStackHCILoadBalancer, vm)
+			if err := r.Client.Update(clusterScope.Context, vm); err != nil {
```
pradipd, line 136: Why do we have to do an update before delete?

pradipd: That seems like an odd pattern.  update an object, just so you can delete it.

lpdnju: Because we need to let controller of vm log the updated correlationId  when it reconcile the VM deletion. e.g when customer deletes loadbalancer, the new correlationId will be generated by RP and updated to loadbalancer CR. We need pass that new correlationId to downstream VM CR to let VM controller logging it.

lpdnju: I tried to find if a better way that can combine update&delete in one client call but not found

pradipd: i see that we use this pattern all over the code.  so, i have approved the PR.

However, i would still like to understand how this works.
So, you have update the objects annotation with the correlation id.  but, then you just delete it.
how would it get logged in telemetry?

pradipd (approved):

controllers/azurestackhciloadbalancer_virtualmachine.go
lpdnju: VM object will not be deleted directly because it has finalizer. a finalizer causes “delete” on the object to become an “update” to set deletion timestamp. 
When VM controller reconciles the VM object, it 
  1:  adds updated correlationId in logger firstly. Then it finds the object has deletiontimestamp which means the object should be deleted. 
  2: call moc to delete VM in hyper-V 
  3. remove finalizer if moc deletes VM successfully. After this, the CR object will be deleted in etcd finally.

pradipd: right.  forgot about the finalizer.

PR #250, opened 2024-01-22
AzureStackHCIVirtualMachine Deletion Bug fix, tests, mocks and fixes for issues including linting, codeql, tsa, etc

PavanNeerudu:
## AzureStackHCIVirtualMachine Deletion Bug fix, tests, mocks and fixes for issues including linting, codeql, tsa, etc
The initial goal of the PR was to fix an issue in the reconcile deletion loop for azurestackhcivirtualmachine(`controllers/azurestackhcimachine_controller.go`). The deletion of azurestackhcivirtualmachine was being handled in a bit wrong way where the deltion of this VM was assumed as a synchronous operation which is not true. This is because, when we issue a delete to a custom resource, it has to go through the logic in the reconciliation loop for this specific resource and post that the resource is deleted. This PR updates the implementation by doing the following:
* If the Get call for the VM results in an err saying that the vm is not present, then we are good(the machine is already deleted).
* If the Get call for the VM results in other errors, the request is reconciled immediately(a non nil error is send)
* If we are able to get the machine and the deletion timestamp on the machine is not zero(the delete was issue, but the deletion did not complete), then the resource is requeued after a short amount of time. However, if the deletion timestamp is not nil, we go ahead and issue delete for this resource and requeue post successful operation.

### Along, with fixing this bug, this PR does the following:
- Adds **initial testing framework**  with gingko based tests and mocks.
- Adds ginkgo based tests covering different scenarios and edge cases for the PR change specific funtion(`reconcileVirtualMachineDelete`) by using a mock client generated using mockgen for controller-runtime's client.
image
- Fixes linting issues(some bugs in the code were found and addressed) by calling lint target in the test and build recipes. Removes the unused cloudtest package. Skipped linting using markers when not sure of the dependencies downstream.
- Fixes kustomization.yaml by replacing the deprecated feature(`patchesStrategicMerge`). `kustomize edit fix` was recursively run against the `kustomization.yaml` files.
- Removes external docker image references for the image used in the Dockerfile(replaced with cbl mariner's distroless minimal image), kube-rbac-proxy, etc.
image
- Fixes CodeQL and TSA issues failing on the master.
image
- Removes the unused versions `v1alpha3` and `v1alpha4` from the controller-gen calls and updates these CRDs.
- Enhancements to make recipes of test and manager builder by calling linting, fmt, vet, module cleaning, etc.
- Pipeline build stage enhancements by calling the test recipe as well as a stage to make sure the go generation doesn't generate any excess files/changes than the ones checked in.


pradipd:
This is great, but, can we break this into 2 pr's?  
1. the fix.
2. the rest of the test framework

If we need to cherry pick the fix for this PR, it would be very challenging with all the test fraweowkr stuff.

PavanNeerudu:
> This is great, but, can we break this into 2 pr's?
> 
> 1. the fix.
> 2. the rest of the test framework
> 
> If we need to cherry pick the fix for this PR, it would be very challenging with all the test fraweowkr stuff.

Hi Pradip. The changes made are kind of reliant on each other(other than things like kustomization, crds, tsa, etc). Do you think once the changes in this PR are approved, can we raise a PR to the release branch with the changes that are actually needed for the fix(just the azurestackhcimachine controller changes).

controllers/azurestackhcimachine_controller.go
```diff
@@ -381,56 +381,61 @@ func (r *AzureStackHCIMachineReconciler) reconcileVirtualMachineNormal(machineSc
...
-			if !apierrors.IsNotFound(err) {
-				return errors.Wrapf(err, "failed to delete AzureStackHCIVirtualMachine %s", vmName)
-			}
+			return reconcile.Result{}, errors.Wrapf(err, "failed to get AzureStackHCIVirtualMachine %s", vmName)
 		}
+		// If the VM resource is not found, no need to reconcile again
+		return reconcile.Result{}, nil
+	}
+
+	// If the VM resource exists and has a deletion timestamp, it means a deletion has been requested.
```
zawachte, line 411: Is there any harm in issuing a delete call again here?

controllers/azurestackhcimachine_controller.go
```diff
@@ -381,56 +381,61 @@ func (r *AzureStackHCIMachineReconciler) reconcileVirtualMachineNormal(machineSc
...
+	// In this case, requeue the request after a delay to check again later if the deletion has been completed.
+	if !vm.DeletionTimestamp.IsZero() {
+		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
```
zawachte, line 414: should we log in this case?

config/crd/kustomization.yaml
```diff
@@ -1,8 +1,6 @@
 # This kustomization.yaml is not intended to be run by itself,
 # since it depends on service name and namespace that are out of this kustomize package.
 # It should be run by config/
-commonLabels:
```
zawachte, line 4: hoping this get functionality is still kept.

controllers/azurestackhcimachine_controller.go
PavanNeerudu: Yeah, good point.

config/crd/kustomization.yaml
PavanNeerudu: kustomization files were autoedited when `kustomization edit fix` was run recursively against the repository. So, yeah, the functionality should be kept intact.

controllers/azurestackhcimachine_controller.go
PavanNeerudu: If you were to issue another delete call here, it would likely be redundant because the DeletionTimestamp being non-zero indicates that a delete request has already been issued. What say? do we want to do it again here?

PavanNeerudu: I was looking online for this, and I could see that while there is no harm to do that again as k8s can handle idempotent operations, its best to avoid doing it again.

controllers/azurestackhcimachine_controller.go
PavanNeerudu: Added additional logs in the last commit. Thanks.

zawachte (approved):

PR #253, opened 2024-01-27
Add error reason for control plane out of range.

mchen727:
Verify that the LB VIP in within the VIP pool range. 

Propagates an error through `LoadBalancerInfrastructureReadyCondition` if the ip is out of range via `LoadBalancerVIPOutOfRangeReason`.


controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -370,3 +380,39 @@ func (r *AzureStackHCILoadBalancerReconciler) reconcileStatus(lbs *scope.LoadBal
 
 	lbs.SetPhase(infrav1.AzureStackHCILoadBalancerPhaseProvisioned)
 }
+
+func (r *AzureStackHCILoadBalancerReconciler) reconcileLoadBalancerVIP(lbs *scope.LoadBalancerScope, clusterScope *scope.ClusterScope) error {
```
madhanrm, line 384: can we find out what would happening when Arc Vnets is used

controllers/azurestackhciloadbalancer_controller.go
```diff
@@ -223,6 +226,13 @@ func (r *AzureStackHCILoadBalancerReconciler) reconcileNormal(lbs *scope.LoadBal
 		}
 	}
 
+	err = r.reconcileLoadBalancerVIP(lbs, clusterScope)
```
aeweston-msft, line 229: Can we move this check up above reconcileVirtualMachines?

Rationale is that if the spec is not valid we would not want to create any child resources. We could also put this into a webhook but I think having it in the reconciler is OK in this case since it has to do a more complex operation to validate (interacting with MOC) with several calls that could theoretically fail, and webhooks have a limited interface to report retryable errors vs. spec content errors.

mchen727: I added the reconcile helper below the LB reconcile because I was using the Address() helper to get the IP. Im happy to move it before reconcileVirtualMachines. Is there a better way to get the IP?

aeweston-msft: OK interesting, I did not realize we were waiting for the IP address assignment to happen and be updated on the underlying MOC resource. How does the IP address of the LB get assigned? If MOC is assigning it and CAPH is just picking it up, how can it be outside of the valid range?

I was assuming that the LB controller was the one setting the LB VIP and that data was passed to it as part of the LB CR and we were just adding a validation for it, but that is not correct, so I am trying to understand the flow here. It looks like in the LB cloud service the MOC LB resource is created providing the vnet name from spec in the FrontendIPConfigurations[0].Subnet.ID field, and then the LB CR status Address field you are checking against is coming from ((*lb.FrontendIPConfigurations)[0].IPAddress).

madhanrm:
Auto-closing this PR as part of the monthly stale-PR cleanup - it has been open for over 90 days or has had no activity in the last 90 days. This is reversible: reactivate the PR if it is still needed. - automated cleanup on behalf of @madhanm

PR #254, opened 2024-02-06
Go Version 1.20 Bump

martiniadam:
This is to prepare for the CAPI 1.5 bump, we need to have Go 1.20 to be able to run it


martiniadam:
Covered in another PR

PR #259, opened 2024-02-25
Remove old APIs to help with removing older module dependencies.

PavanNeerudu:
This PR removes old api versions v1alpha3 and v1alpha4 from the repository. These versions are old and are no longer used in any deployment.
When the cluster-api module is bumped in mgmtappl, the current CAPH code v1alpha3 and v1alpha4 api's import older api's of cluster-api which are removed in the newer versions. This is causing some problems with modules upgrade in other upstream repositories. Also, when the cluster-api module is updated to higher version, CAPH is also going to hit similar problems.


zawachte (approved):

zawachte:
Thanks for this!

PR #261, opened 2024-03-11
Don't wrap not-found error

nistopia:
We were wrapping ResourceNotFound error in Get call but checking for ResourceNotFound error after calling Get. We don't need to wrap ResourceNotFound  to include the resource name as ResourceNotFound error contains the resource name.


cloud/services/disks/disks.go
```diff
@@ -40,7 +40,7 @@ func (s *Service) Get(ctx context.Context, spec interface{}) (interface{}, error
 	}
 	disk, err := s.Client.Get(ctx, s.Scope.GetResourceGroup(), "", diskSpec.Name)
 	if err != nil && azurestackhci.ResourceNotFound(err) {
-		return nil, errors.Wrapf(err, "disk %s not found", diskSpec.Name)
+		return nil, err
```
pradipd, line 43: This can be collapsed down to 
```
if err != nil {
  return nil, err
}
```

cloud/services/vippools/vippools.go
```diff
@@ -41,7 +41,7 @@ func (s *Service) Get(ctx context.Context, spec interface{}) (interface{}, error
 
 	vp, err := s.Client.Get(ctx, vpSpec.Location, vpSpec.Name)
 	if err != nil && azhci.ResourceNotFound(err) {
-		return nil, errors.Wrapf(err, "vippool %s not found", vpSpec.Name)
+		return nil, err
```
pradipd, line 44: this can be collapsed down to
```
if err != nil {
  return nil, err
}

File: cloud/services/disks/disks.go [hunk 3]
```diff
@@ -40,7 +40,7 @@ func (s *Service) Get(ctx context.Context, spec interface{}) (interface{}, error
 	}
 	disk, err := s.Client.Get(ctx, s.Scope.GetResourceGroup(), "", diskSpec.Name)
 	if err != nil && azurestackhci.ResourceNotFound(err) {
-		return nil, errors.Wrapf(err, "disk %s not found", diskSpec.Name)
```
pradipd, line 43: Does MOC provide enough information in its error messages?

cloud/services/disks/disks.go
nistopia: Updated.

cloud/services/vippools/vippools.go
nistopia: Done.

cloud/services/disks/disks.go
nistopia: Yes, it does. It includes the name of the resource it didn't find.

zawachte (approved):

PR #266, opened 2024-03-18
Check CloudAgent reachability in caph liveness probe

PavanNeerudu:
### Description
The current implementation of liveness probe uses controller-runtime package's healthz.Ping which returns no error by default. So as long as the manager runs, the check returns a successs.

This PR updates the implementation by using a new method EndPointChecker instead of healthz.Ping. The new method takes the endpoint as argument and performs a tcp check with exponential backoff against that endpoint. The values for the exponential backoff are obtained from the diagnostics-checker and the cloudagent fqdn is obtained from the environment variable which is then appended with 55000 as port.

### Testing

- Positive check result: Ensured that there are no container restarts when the cloudagent is running.
- Negative check result: Manually stopped the cloudagent service, and saw the container getting restarted. The following is a snapshot from the when `kubectl describe` was run for caph pod:
  image
Snapshot of the caph pod logs:
image


cmd/manager/main.go
```diff
@@ -253,19 +255,32 @@ func main() {
...
+	cloudAgentFqdn := os.Getenv("AZURESTACKHCI_CLOUDAGENT_FQDN")
+	if cloudAgentFqdn == "" {
+		setupLog.Error(nil, "AZURESTACKHCI_CLOUDAGENT_FQDN is not set")
 		os.Exit(1)
 	}
 
-	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
+	cloudAgentAddress := cloudAgentFqdn + ":55000"
```
zawachte, line 281: Should we be hard coding this port? How does the rest of the get this?

pkg/network/network.go
```diff
@@ -0,0 +1,40 @@
...
+
+		backoff := wait.Backoff{
+			Duration: 250 * time.Millisecond,
+			Factor:   1.5,
+			Steps:    9,
+			Jitter:   0.1,
+		}
+		defaultTimeout := 1 * time.Second
+
+		err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
zawachte, line 25: Did we consider an approach that actually uses the client grpc moc lib to do some sort of check?

TBH ... I think the most useful approach would be:

1. On startup create a valid session/login with moc.
2. Pass the context into the liveness probe context.
3. Attempt a moc GET each probe.
4. If the GET valids ... fail the probe.

Thoughts?

PavanNeerudu: Yes, my initial thought process was to handle this similarly to how we test MOC connectivity during the early stages of kva deployment. However, inside the CAPH, the code attempts to reconfigure the MOC login by checking if an env variable `WSSD_CONFIG_PATH` is present. If this variable is absent, it tries to acquire the login config from a secret `caphlogintoken` created by the cloud-operator. It accesses the data from this secret, and makes subsequent MOC calls to get the authorizer. Its just that, There is a lot of logic in the authentication happening here, https://github.com/microsoft/cluster-api-provider-azurestackhci/blob/master/pkg/auth/auth.go and I wanted to keep this check simpler. 
If we go by this method, it requires us to get the secret in the described manner (since I haven't found an environment variable named WSSD_CONFIG_PATH in the CAPH deployment, and we're not mounting the login config). We would need to obtain the manager client before starting the manager and pass this client to the check, and call those auth package functions to get us the authorizer to be used in the MOC get calls. Alternatively, I think we should be able to update the caph deployment to include this env and mount the login config to the pod(but yeah, we would need the upstream callers to create the file and mount the path to the pod)

cmd/manager/main.go
PavanNeerudu: I don't see anywhere we use this Port. 
For day0 kva deployment, we take the cloudfqdn and port information(defaulted to 55000) and just use this info for the diagnostics-checks. While we create the CloudInfrastruture CRD with this data, cloud-operator is never using this info.

pkg/network/network.go
zawachte: Thanks for the breakdown and investigation.

So it sounds like this PR is just a starting point and we can make incremental progress on this?

cmd/manager/main.go
zawachte: I meant like is this portinfo passed into CAPH?

I guess it is by the auth secret that we spoke about on the other thread. 

I just worry about needing to support cloudagent with a custom port one day and then this liveness probe breaks.

zawachte:
We may also need to tweak the default liveness probe parameters in the deployment spec. Thoughts?

cmd/manager/main.go
PavanNeerudu: caph is not being passed any port through environment variables/other secrets. And for supporting the custom port, we need changes from the upstream cloud-operator. 
cloud-operator is taking the input but not passing/using this port info.
image

We can make it pass the port details to caph, just like how we are passing the fqdn details but it doesn't use anywhere.

pkg/network/network.go
PavanNeerudu: Yeah, my thoughts as well. Wanted to keep it simple for the initial pass and we make an incremental progress on this.

PavanNeerudu:
> We may also need to tweak the default liveness probe parameters in the deployment spec. Thoughts?

Yeah, that would be a better solution. Instead of modifying any deployment spec, can we pass in these as arguments to the manager binary by defaulting it to certain values. This would be simple and does not break deployments as well.

PR #272, opened 2024-04-10
Migration to MCR

martiniadam:
We are publishing our images to MCR, and part of doing that is changing our pipelines to push to where MCR can pick it up. This PR:

- Changes the kustomize templates and pipelines to go to this webhook.

Still open question: 
- Are we changing staging portion to get the internal hook? Or should we keep it on mocimages?

### DIFF
--- a/templates/flavors/base/cluster-template.yaml
+++ b/templates/flavors/base/cluster-template.yaml
@@ -326,7 +326,7 @@ spec:
           bind-address: "0.0.0.0"
           leader-elect-lease-duration: "60s"
           leader-elect-renew-deadline: "55s"
-      imageRepository: "ecpacr.azurecr.io"
+      imageRepository: "mcr.microsoft.com"
   version: "${KUBERNETES_VERSION}"
 ---
 apiVersion: cluster.x-k8s.io/v1beta1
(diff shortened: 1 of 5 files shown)


templates/flavors/base/cluster-template.yaml
```diff
@@ -326,7 +326,7 @@ spec:
           bind-address: "0.0.0.0"
           leader-elect-lease-duration: "60s"
           leader-elect-renew-deadline: "55s"
-      imageRepository: "ecpacr.azurecr.io"
```
zawachte, line 329: this is unrelated. Please revert.

templates/flavors/mgmt/mgmt-machine.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         leader-elect-lease-duration: 60s
         leader-elect-renew-deadline: 55s
         terminated-pod-gc-threshold: "10"
-    imageRepository: ecpacr.azurecr.io
```
zawachte, line 55: this is unrelated. Please revert.

martiniadam: reverted

templates/flavors/base/cluster-template.yaml
martiniadam: reverted

PR #281, opened 2024-08-21
wait for volume attachments to be detached before deleting a vm

georgel-ms:
This is to fix a bug that during workload cluster upgrade, it takes long time to wait for the PV to attach to the new node, we may see multiple errors like this:
 
I0404 11:45:49.477443       1 event.go:307] "Event occurred" object="default/sample-dddf7bd5-wglbx" fieldPath="" kind="Pod" apiVersion="v1" type="Warning" reason="FailedAttachVolume" message="Multi-Attach error for volume \"pvc-d0cec9c7-7d3a-4961-ab35-7bda48cbe5ee\" Volume is already used by pod(s) sample-dddf7bd5-7rh7q"

In this PR, we intentionally wait for the PV to detach from the old node before the vm is deleted to avoid the above error.

### DIFF
--- a/cloud/scope/virtualmachine.go
+++ b/cloud/scope/virtualmachine.go
@@ -139,6 +139,10 @@ func (m *VirtualMachineScope) ClusterName() string {
 	return m.AzureStackHCIVirtualMachine.Spec.ClusterName
 }
 
+func (m *VirtualMachineScope) Client() client.Client {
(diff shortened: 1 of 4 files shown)


zawachte:
CAPI is already supposed to wait for the PVs to be detached before scheduling node delete.

PR #285, opened 2024-10-25
Adding call to MOC to remove ISO when machines is healthy

pradipd:
Removing the ISO when the machine is marked healthy.


madhanrm (approved):

zawachte:
@pradipd I left some comments on a previous draft. Can we discuss them?

controllers/azurestackhcivirtualmachine_controller.go
```diff
@@ -175,6 +189,14 @@ func (r *AzureStackHCIVirtualMachineReconciler) reconcileNormal(virtualMachineSc
 		virtualMachineScope.Info("Machine VM is running", "name", virtualMachineScope.Name())
 		virtualMachineScope.SetReady()
 		conditions.MarkTrue(virtualMachineScope.AzureStackHCIVirtualMachine, infrav1.VMRunningCondition)
+
+		if conditions.IsTrue(virtualMachineScope.Machine, clusterv1.MachineNodeHealthyCondition) {
+			err = ams.removeISODisk()
```
pradipd, line 194: From Zach:
Do you know if we have any other cases in the code today where we are depending on the status of a core provider resource status to make a decision in the infra provider?
I am trying to think of any case where this could go wrong or create some sort of circular dependency, but nothing super super obvious is coming to mind.

controllers/azurestackhcivirtualmachine_controller.go
```diff
@@ -175,6 +189,14 @@ func (r *AzureStackHCIVirtualMachineReconciler) reconcileNormal(virtualMachineSc
...
+			if err != nil {
```
pradipd, line 195: From Zach:
More of a note to myself to see if MachineNodeHealthyCondition or just readycondition is better to key off of here.

controllers/azurestackhcivirtualmachine_controller.go
pradipd: @zawachte  FYI, I moved your comment into this PR.

controllers/azurestackhcivirtualmachine_controller.go
pradipd: @zawachte FYI I moved your comment into this PR.

pradipd:
> @pradipd I left some comments on a previous draft. Can we discuss them?

FYI, I moved the comments into this PR (the other was just a commit) so we can discuss.

pradipd:
Ooops. i didn't mean to close that.  I hit the wrong button.

controllers/azurestackhcivirtualmachine_controller.go
pradipd: I choose MachineNodeHealthCondition because it comes later than Ready.

```
- lastTransitionTime: "2024-10-25T18:37:44Z"
  status: "True"
  type: BootstrapReady
- lastTransitionTime: "2024-10-25T18:38:44Z"
  status: "True"
  type: Ready
- lastTransitionTime: "2024-10-25T18:38:44Z"
  status: "True"
  type: InfrastructureReady
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: APIServerPodHealthy
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: ControllerManagerPodHealthy
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: EtcdMemberHealthy
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: EtcdPodHealthy
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: HealthCheckSucceeded
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: NodeHealthy
- lastTransitionTime: "2024-10-25T18:39:15Z"
  status: "True"
  type: SchedulerPodHealthy

```

pradipd: I'm looking around to see what "ready" means.

controllers/azurestackhcivirtualmachine_controller.go
pradipd: i do not know offhand, but, will look.

pradipd: I do not see any cases in CAPH where we are depending on the status of a core provider.

pradipd: Are you thinking it would be better to drive this from the upper layers?
E.g. once the cluster is created, cloud-op can update the machine* specs and then have azstackhci*machine can  reconcile it and treat it as an update?
Or something like that?
I think this is more correct controller pattern.

PR #291, opened 2025-02-25
Pipeline fix: bump golangci and linters

martiniadam:
Remove golangci, linters, and ignore new checks for G115: overflow conversion to not change the API. Removing dead linters that aren't being used.


martiniadam:
Successful daily run: https://github-private.visualstudio.com/microsoft/_build/results?buildId=94417&view=results

zawachte (approved):

PR #294, opened 2025-06-05
Add Generic Not Found case

mchen727:
Not Found error is considered a Legacy error code and does not get picked up by GetErrorCode()
GetErrorCode() returns NotFound error for PathNotFound, BlobNotFound, FileNotFound and generic NotFound

Also adding a generic error message to catch generic NotFounds.

### DIFF
--- a/api/v1beta1/conditions_consts.go
+++ b/api/v1beta1/conditions_consts.go
@@ -38,6 +38,8 @@ const (
 	OutOfNodeCapacityReason = "OutOfNodeCapacity"
 	// PathNotFoundReason used when the AzureStackHCI GalleryImage is not found.
 	PathNotFoundReason = "PathNotFound"
+	// NotFoundReason used as a generic error reason when a resource is not found.
+	NotFoundReason = "PathNotFound"
 )
 
 // Conditions and condition Reasons for the AzureStackHCICluster object
--- a/api/v1beta1/conditions_consts.go
+++ b/api/v1beta1/conditions_consts.go
@@ -39,7 +39,7 @@ const (
 	// PathNotFoundReason used when the AzureStackHCI GalleryImage is not found.
 	PathNotFoundReason = "PathNotFound"
 	// NotFoundReason used as a generic error reason when a resource is not found.
-	NotFoundReason = "PathNotFound"
+	NotFoundReason = "NotFound"
 )
 
 // Conditions and condition Reasons for the AzureStackHCICluster object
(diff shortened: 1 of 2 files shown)


rjaini (dismissed):

api/v1beta1/conditions_consts.go
```diff
@@ -38,6 +38,8 @@ const (
 	OutOfNodeCapacityReason = "OutOfNodeCapacity"
 	// PathNotFoundReason used when the AzureStackHCI GalleryImage is not found.
 	PathNotFoundReason = "PathNotFound"
+	// NotFoundReason used as a generic error reason when a resource is not found.
+	NotFoundReason = "PathNotFound"
```
pradipd, line 42: Should this be "PathNotFound"?  that looks odd.
i would guess that it should be:
NotFoundReason = "NotFound"

controllers/azurestackhcivirtualmachine_controller.go
```diff
@@ -210,8 +211,13 @@ func (r *AzureStackHCIVirtualMachineReconciler) getOrCreate(virtualMachineScope
 				conditions.MarkFalse(virtualMachineScope.AzureStackHCIVirtualMachine, infrav1.VMRunningCondition, infrav1.OutOfCapacityReason, clusterv1.ConditionSeverityError, err.Error())
 			case mocerrors.OutOfNodeCapacity.Error():
 				conditions.MarkFalse(virtualMachineScope.AzureStackHCIVirtualMachine, infrav1.VMRunningCondition, infrav1.OutOfNodeCapacityReason, clusterv1.ConditionSeverityWarning, err.Error())
-			case mocerrors.PathNotFound.Error():
-				conditions.MarkFalse(virtualMachineScope.AzureStackHCIVirtualMachine, infrav1.VMRunningCondition, infrav1.PathNotFoundReason, clusterv1.ConditionSeverityError, err.Error())
+			// Internally, NotFound is a legacy error and returns the error string instead.
+			case moccodes.NotFound.String():
```
pradipd, line 215: i'm a little confused.
Why are we checking moccodes here, but mocerrors everywhere else?

api/v1beta1/conditions_consts.go
pradipd: Perhaps, "ObjectNotFound".  
It could be "MocObjectNotFound".  but, i don't know if people will object to exposing underlying implementation details as an end user error.

controllers/azurestackhcivirtualmachine_controller.go
pradipd: Based off your description i would think it should be somethinig like:
case mocerrors.NotFound.Error():
    if <path not found>
        //set error condition to PathNotFound
    else
        //set error condition to NotFound

mchen727: During my testing I found that GetErrorCode(err) was returning 'NotFound' for the error code. Underneath, we are grouping the various NotFound errors into a single NotFound case. https://github.com/microsoft/moc/blob/6a885e8c5cea2264a980f6ce79ee3c1246115711/pkg/errors/errors.go#L227

However, mocerrors.NotFound.Error() returns 'Not Found' (with a space) because it is classified as a legacy error code. I elected the moccodes.NotFound.String() to properly match the case,

pradipd: would it make sense to handle both of them then together?
case mocerrors.NotFound:  //"Not Found"
   fallthrough
case moccodes.NotFound: //"NotFound"

   if mocerrors.IsPathNotFound(err) {
       //set condition
   } else {
   }

pradipd (approved):

PR #296, opened 2025-06-17
Remove unnecessary disk service and delete step

aeweston-msft:
# Changes introduced by this PR
* Remove the disk service from AzureStackHCIVirtualMachineService struct and the delete step from AzureStackHCIVirtualMachineService Delete flow
* Remove the entire disk service which wraps moc-sdk-for-go, as it is unused after the change

I have been working on some task which required a review of the CAPH interactions with MOC and I realized that there is no need for the OS disk delete step.
 
This is because the OS disk is a resource whose lifecycle is completely managed within the VM service of MOC and is not represented as a VHD (It has a separate resource called VM image which is only referenced internally to MOC AFAIK). So this is essentially a no-op because this operation should always result in not found. I thought it would be good to remove this code and the corresponding disk service entirely to avoid having dead code and unnecessary calls to MOC. I was just wondering before making a PR what the testing expectations are for the CAPH repo. 

# Test plan
TBD, unsure of the expectations


madhanrm:
Auto-closing this PR as part of the monthly stale-PR cleanup - it has been open for over 90 days or has had no activity in the last 90 days. This is reversible: reactivate the PR if it is still needed. - automated cleanup on behalf of @madhanm

PR #299, opened 2025-10-08
CAPI 1.11

martiniadam:
This PR updates CAPH to use CAPI 1.11, and updates the provider to use v1beta2 CAPI API's

Please note that this will bring up CAPH to v1.2.0


martiniadam:
Successful run of KVA pipelines with dev image (as well as cloud-operator, mgmtappl changes) https://dev.azure.com/msazure/msk8s/_build/results?buildId=153302557&view=results

controllers/azurestackhcicluster_controller.go
```diff
@@ -157,11 +157,29 @@ func (r *AzureStackHCIClusterReconciler) reconcileNormal(clusterScope *scope.Clu
...
+				Reason:  infrav1.OutOfMemoryReason,
+				Message: err.Error(),
+			}
+			conditions.Set(azureStackHCICluster, notReady)
 		case mocerrors.OutOfCapacity.Error():
-			conditions.MarkFalse(azureStackHCICluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.OutOfCapacityReason, clusterv1.ConditionSeverityError, err.Error())
+			outOfCapacityCondition := metav1.Condition{
+				Type:    infrav1.NetworkInfrastructureReadyCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  infrav1.OutOfMemoryReason,
```
nwoodmsft: Why OutOfMemoryReason is returned for OutOfCapacity case?

controllers/azurestackhcimachine_controller.go
```diff
@@ -207,11 +210,28 @@ func (r *AzureStackHCIMachineReconciler) reconcileNormal(machineScope *scope.Mac
...
-	if !machineScope.Cluster.Status.InfrastructureReady {
+	// Check if the infrastructure cluster is ready by checking our own AzureStackHCICluster status
+	if clusterScope.AzureStackHCICluster.Status.Initialization == nil ||
+		clusterScope.AzureStackHCICluster.Status.Initialization.Provisioned == nil ||
+		!*clusterScope.AzureStackHCICluster.Status.Initialization.Provisioned {
 		machineScope.Info("Cluster infrastructure is not ready yet")
 		return reconcile.Result{}, nil
 	}
 
+	// Set machine infrastructure as provisioned so CAPI knows we're ready to accept bootstrap data
```
nwoodmsft: On master, Status.Ready was only set after VMStateSucceeded (old line 253). The new code at lines 221-231 would set Initialization.Provisioned = true before the VM is even created by MOC, along with an
  immediate PatchObject() to persist it. That seems strange?

It looks like Initialization.Provisioned is just v1beta2's equivalent of the old Status.Ready. CAPI should populate 
  Machine.Spec.Bootstrap.DataSecretName independently (the old code checked for it at line 216 with no dependency on Ready). 

SetReady() is called on line 278 and that should set Provisioned = true correctly...so I am not sure if we should be handling this in the line 226-233 block...

controllers/azurestackhcicluster_controller.go
```diff
@@ -233,14 +292,21 @@ func (r *AzureStackHCIClusterReconciler) reconcileDelete(clusterScope *scope.Clu
...
+			Reason: infrav1.LoadBalancerDeletingReason,
+		})
 		return reconcile.Result{RequeueAfter: 20 * time.Second}, nil
 	}
 
 	if err := newAzureStackHCIClusterReconciler(clusterScope).Delete(); err != nil {
 		wrappedErr := errors.Wrapf(err, "error deleting AzureStackHCICluster %s/%s", azureStackHCICluster.Namespace, azureStackHCICluster.Name)
 		r.Recorder.Eventf(azureStackHCICluster, corev1.EventTypeWarning, "FailureClusterDelete", wrappedErr.Error())
-		conditions.MarkFalse(azureStackHCICluster, infrav1.NetworkInfrastructureReadyCondition, clusterv1.DeletionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
+		conditions.Set(azureStackHCICluster, metav1.Condition{
```
nwoodmsft, line 306: Missing Status field in metav1.Condition ?

metav1.Condition.Status is +required with +kubebuilder:validation:Enum=True;False;Unknown — the Kubernetes API definition explicitly marks it as required and validates it must be one of those three values. An
   empty string "" is invalid.

All 4 other "DeletionFailed" conditions in this file correctly include Status: metav1.ConditionFalse

controllers/azurestackhcimachine_controller.go
nwoodmsft: Let's also compare to the behavior of the azureedgemachine_controller.go in mono repo (it doesn't seem to set initialization.provisioned until the end of reconcilation, after the bootstrap data is available).

controllers/azurestackhcicluster_controller.go
martiniadam: Yes thank you, fixed.

controllers/azurestackhcicluster_controller.go
martiniadam: A mistake, reason is now OutOfCapacity

controllers/azurestackhcimachine_controller.go
martiniadam: That makes sense, I may have added this as a workaround for some deploy issues I was having before. Fixed and retesting.

martiniadam:
Successful run after review changes (one test flake) https://msazure.visualstudio.com/msk8s/_build/results?buildId=153681433&view=results

nwoodmsft (approved):

PR #300, opened 2025-10-22
Add oras EOL annotation step to build

cindyxmiao:
- Add similar EOL annotation script we have for other repos to CAPH (but in-line script)
- Also download oras
- For balance of speed/functionality purposes, fetches latest 20 tags (excluding the current tag), checks if they are annotated, and marks them as expired iff they are not annotated already.


cindyxmiao:
@microsoft-github-policy-service agree company="Microsoft"

cindyxmiao:
pipeline on azure

PR #308, opened 2026-04-22
fix: handle empty Reason in v1beta1-to-v1beta2 condition and OSDisk conversion

madhanrm:
## Problem

After upgrading CAPI/CAPH from v1beta1 to v1beta2 storage, **all status patches fail with 422 Unprocessable Entity**:

```
conditions[0].reason: Invalid value "": conditions[0].reason in body should be at least 1 chars long
```

This blocks ALL machine provisioning after an upgrade — 9,407 occurrences observed in a single pipeline run.

### Root Cause

1. Objects created by old CAPH (v1beta1 API) store conditions where `Reason` is empty string
2. v1beta1 `Reason` field has `json:"reason,omitempty"` — empty string is **omitted** from stored JSON — valid
3. After CAPI/CAPH upgrade, v1beta2 CRD is registered with `reason: required, minLength: 1`
4. On read, API server converts v1beta1 to v1beta2: `out.Reason = in.Reason` copies empty string
5. v1beta2 JSON serializes `"reason": ""` (no omitempty) — fails CRD validation
6. Every status patch rejected — `provisioned` stays false forever — pipeline times out

### Evidence

| Machine | Created By | Status Patch |
|---------|-----------|--------------|
| pgdzn (new mgmt CP) | New CAPH (v1beta2) | Succeeds |
| control-plane-0 (old mgmt CP) | Old CAPH (v1beta1) | 422 (24x) |
| workload-control-plane-cv9gn | Old CAPH (v1beta1) | 422 (143x) |
| testnodepool-* (4 machines) | Old CAPH (v1beta1) | 422 (128x each) |

### Fix

**Condition conversion** (line 161): Default empty Reason to the condition Type during v1beta1 to v1beta2 conversion. This satisfies CRD validation while preserving semantic meaning.

**OSDisk conversion** (line 175): Always create ManagedDisk pointer since v1beta2 CRD requires it. Previously, empty StorageAccountType left ManagedDisk nil, causing `bastion.osDisk.managedDisk: Required value`.

### Testing

- [x] `go build ./...` passes
- [ ] Integration test with upgrade pipeline

### DIFF
--- a/api/v1beta1/conversion.go
+++ b/api/v1beta1/conversion.go
@@ -159,6 +159,12 @@ func Convert_v1beta1_Condition_To_v1_Condition(in *corev1beta1.Condition, out *m
 	out.Status = metav1.ConditionStatus(in.Status)
 	out.LastTransitionTime = in.LastTransitionTime
 	out.Reason = in.Reason
+	// v1beta1 Reason is optional (json omitempty), but v1beta2 metav1.Condition
+	// requires Reason with minLength=1. Objects stored as v1beta1 may have empty
+	// Reason, which would fail CRD validation after upgrade to v1beta2 storage.
+	if out.Reason == "" {
+		out.Reason = string(in.Type)
+	}
 	out.Message = in.Message
 	// corev1beta1.Condition doesn't have ObservedGeneration field
 	return nil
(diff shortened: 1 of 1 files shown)


martiniadam:
@madhanrm Can we see if Message is also handled? We hit that same validation on going to 1.12 in cloud-op, want to get ahead of that if that's something I missed

madhanrm:
@martiniadam please take this forward and fix this.

If there were a CI on this, I would have gotten this validated. I dont have cycles to manually try and validate this.

PR #315, opened 2026-07-07
Surface terminal VM provisioning failures on the CAPI contract Ready condition

nistopia:
## Summary

When a VM `Create()` fails terminally during a nodepool surge / rolling node replacement (capacity, `InvalidVMSku`, missing VHD path, etc.), CAPH classifies the error and goes terminal, but only records it on **`VMRunningCondition`**. CAPI v1.12.5 (`setInfrastructureReadyCondition`) mirrors the infra machine's **contract `"Ready"` condition** into `Machine.InfrastructureReadyCondition`. Since CAPH never set a `"Ready"` condition, CAPI fell back to the generic `"<Kind> status.initialization.provisioned is false"` message and the **real SKU/capacity error was dropped** — the upgrade/replacement stalled with an empty `"Detailed message: : Timed out"` and retried for days.

## Fix

Set the CAPI contract `"Ready"` condition (Type `"Ready"`) alongside `VMRunningCondition` for the terminal `Create()` failure cases (OutOfMemory, OutOfCapacity, OutOfNodeCapacity, PathNotFound, NotFound, VMProvisionFailed), carrying the **same reason + message**. The `AzureStackHCIMachine` controller already copies every VM condition onto the machine, so CAPI now mirrors the real reason/message into `InfrastructureReadyCondition`, which the existing cloud-operator pipe surfaces as an actionable terminal error. On `VMStateSucceeded` the `"Ready"` condition is set `True` so a prior failure is cleared.

No cloud-operator change required. Verified against CAPI v1.12.5 (`ReadyConditionType()` -> `"Ready"`, `NewMirrorConditionFromUnstructured` with generic fallback) and the existing VM-condition copy loop.

## Testing

- `go build ./controllers/...`, `go vet ./controllers/` -- clean.
- New unit test `TestSetVMProvisionFailure` asserts an `OutOfCapacity` failure sets `Ready=False` with the message on the `AzureStackHCIVirtualMachine`.

Fixes AB#38511842 (parent AB#38321574).

### DIFF
--- a/controllers/azurestackhcivirtualmachine_controller.go
+++ b/controllers/azurestackhcivirtualmachine_controller.go
@@ -33,6 +33,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


Ziling-Zhou (dismissed):

mpaul-msft:
Please address merge conflicts.

djshah19 (dismissed):

nwoodmsft (approved):

PR #318, opened 2026-07-27
Bump Go to 1.26.5 and patch grpc + x/text CVEs

mpaul-msft:
Component Governance / vulnerability remediation.

go directive 1.25.11 -> 1.26.5 clears GO-2026-5856 (crypto/tls). govulncheck required go1.25.12; 1.26.5 is the KVA fleet floor, so this aligns CAPH with the rest of the fleet in one step. The exact GoTool@0 pin in .pipelines/build.yaml is bumped to match.

go get google.golang.org/grpc@v1.82.1 clears GO-2026-6061 / GHSA-hrxh-6v49-42gf (gRPC-Go xDS RBAC authorization bypass + HTTP/2 rapid-reset DoS). go get golang.org/x/text@v0.39.0 clears GO-2026-5970 (infinite loop on invalid input in golang.org/x/text).

Verification: go build ./... OK; go vet ./... OK; go mod verify OK; govulncheck ./... -> CLEAN, 0 vulnerabilities. Unit tests unchanged from base (controllers suite needs envtest binaries, absent in this environment both before and after this change).

PR #319, opened 2026-07-27
fix: bump Go toolchain to 1.25.12 for GO-2026-4970

cindyxmiao:
S360 flags the CAPH image for [GO-2026-4970](https://pkg.go.dev/vuln/GO-2026-4970) / CVE-2026-39822 — a Go stdlib bug where `os.Root` follows a symlink outside the root when the final path component is a symlink and the path ends in `/` (e.g. `root.Open("symlink/")`).

It's a toolchain vuln, not a dependency or app-code bug — the fix is just recompiling with a patched Go. CAPH is on the 1.25 line, fixed in `1.25.12`:

- `go.mod` / `Makefile` fallback: `1.25.11` -> `1.25.12`
- both `GoTool@0` pins (`azure-pipelines-release.yml`, `.pipelines/build.yaml`): `1.25.11` -> `1.25.12` — the release build is what produces the shipped `manager` binary the scanner re-scans

Staying on the same 1.25 minor line, so no language/behavior change. CAPH doesn't use `os.Root` in its own code (0 usages), so runtime exposure is nil — this is a compile-hygiene bump to clear the scanner.

Needs an image rebuild+republish (via the release pipeline) for the scanner to re-scan the recompiled binary.

### DIFF
--- a/.pipelines/build.yaml
+++ b/.pipelines/build.yaml
@@ -48,7 +48,7 @@ jobs:
 
       - task: GoTool@0
         inputs:
-          version: "1.25.11"
+          version: "1.25.12"
         displayName: "Install Go"
 
       - script: |-
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ SHELL:=/usr/bin/env bash
 # Go version
 GO_VERSION := $(shell go env GOVERSION | sed "s/[^[:digit:].-]//g")
 ifeq ($(GO_VERSION),)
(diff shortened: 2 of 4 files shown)


cindyxmiao:
Closing in favor of #318 — that one bumps Go to 1.26.5 (the KVA fleet floor) which clears GO-2026-4970 plus a superset of CVEs (crypto/tls, grpc, x/text) in one step.

